PR #10, opened 2017-12-15
Support browsing of favourite tracks, moods, genres and playlists

SERVCUBED:
Also added user's favourite playlists to the playlist list (prefixed with an asterisk).


SERVCUBED:
The remaining flake8 I202 errors are not from my code. Would you like me to fix them? In my opinion fixing them makes the code seem messier.

Issue #15, opened 2020-04-21
Slow browsing into songs list


coolky:
This is my environment:
Manjaro 19.02 x64
python 3.8.2
mopidy 3.0.2
mopidy-tidal 0.2.2

I'm using mopidy-musicbox-webclient, but I have the same issue with all mopidy webclients (and other gtk mpd client). When I'm trying to browse into an album or playlist I have to wait a lot of time until I can choose to play one song, this appens only into tidal browsing.

The logs are crystal clear:

> INFO     2020-04-21 13:13:43,033 [3462:TidalBackend-6] mopidy_tidal.library
  Lookup uris 'tidal:track:6164258:102387249:102387250'
INFO     2020-04-21 13:13:43,761 [3462:TidalBackend-6] mopidy_tidal.library
  Returning 1 tracks
INFO     2020-04-21 13:13:43,761 [3462:TidalBackend-6] mopidy_tidal.library
  Lookup uris 'tidal:track:8149304:86638235:86638238'
INFO     2020-04-21 13:13:44,353 [3462:TidalBackend-6] mopidy_tidal.library
  Returning 1 tracks
INFO     2020-04-21 13:13:44,354 [3462:TidalBackend-6] mopidy_tidal.library
  Lookup uris 'tidal:track:4818468:93636554:93636556'
INFO     2020-04-21 13:13:44,921 [3462:TidalBackend-6] mopidy_tidal.library
  Returning 1 tracks
INFO     2020-04-21 13:13:44,921 [3462:TidalBackend-6] mopidy_tidal.library
  Lookup uris 'tidal:track:9817165:96843255:96843256'
INFO     2020-04-21 13:13:45,530 [3462:TidalBackend-6] mopidy_tidal.library
  Returning 1 tracks
INFO     2020-04-21 13:13:45,530 [3462:TidalBackend-6] mopidy_tidal.library
  Lookup uris 'tidal:track:5701795:93139653:93139654'
INFO     2020-04-21 13:13:46,112 [3462:TidalBackend-6] mopidy_tidal.library
  Returning 1 tracks
INFO     2020-04-21 13:13:46,113 [3462:TidalBackend-6] mopidy_tidal.library
  Lookup uris 'tidal:track:16056:101269540:101269541'
INFO     2020-04-21 13:13:46,716 [3462:TidalBackend-6] mopidy_tidal.library
  Returning 1 tracks
[...]

PR #27, opened 2020-11-21
v0.3.0 OAuth Flow Authentication

quodrum-glas:

PR #30, opened 2021-02-09
implementation of lookup for playlists in library, and track caching

MrSurly:
This is what I hope is a beginning of a fix for #15.  In any case, I talked to the mopidy devs about the issue, and implementing this is probably desirable.

I think a fix for #15 will require changes to Iris.

### DIFF
--- a/mopidy_tidal/library.py
+++ b/mopidy_tidal/library.py
@@ -193,9 +193,16 @@ class TidalLibraryProvider(backend.LibraryProvider):
                 tracks += self._lookup_album(session, parts)
             elif uri.startswith('tidal:artist'):
                 tracks += self._lookup_artist(session, parts)
+            elif uri.startswith('tidal:playlist'):
+                pass
+                tracks += self._lookup_playlist(session, parts)
 
         logger.info("Returning %d tracks", len(tracks))
         return tracks
+
+    def _lookup_playlist(self, session, parts):
+        tracks = session.get_playlist_tracks(parts[2])
+        return full_models_mappers.create_mopidy_tracks(tracks)
 
     def _lookup_track(self, session, parts):
         album_id = parts[3]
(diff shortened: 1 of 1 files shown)


MrSurly:
@tehkillerbee This works for me -- the Mopidy devs recommended adding the track caching from the `lookup` calls.

tehkillerbee:
@MrSurly Great to see you have had progress! Looking forward to test it during the next couple of days.

mopidy_tidal/library.py
```diff
@@ -18,6 +18,24 @@
 
 logger = logging.getLogger(__name__)
 
+class Cache(dict): 
+    '''cache class, requires python3 style (ordered) dicts'''
```
kingosticks, line 22: If this was a an actual `OrderedDict` you could use [`popitem(last=False)`](https://docs.python.org/3.7/library/collections.html#collections.OrderedDict.popitem) to simplify your `__prune()`.

MrSurly: I'm relying on Python 3 dicts (which are essentially `OrderedDict`).  For some reason, I thought Mopidy was Python3 only?

I'll get this changed in any case.

MrSurly: Updated.

kingosticks: You are totally correct that Mopidy requires Python 3.7+ and in Python 3.7, dict became ordered. But dict doesn't have all the same functionality as `OrderedDict`. Popping from the front with `last=False` is only possible with  `OrderedDict`.

MrSurly: I saw that, which is why I changed it.

tehkillerbee:
@MrSurly @kingosticks Nice work, it looks like the caching works as expected - playback starts quite fast now. I still think it would be nice to get some sort of image caching added too (similar to the suggested cache changes in #27), as I believe it would help speeding up the browsing experience too. I will merge these changes in now.

PR #35, opened 2021-04-17
Use new OAuth flow added in upstream tidalapi

greggilbert:
Since the standard login seems to have broken, the upstream [python-tidal](https://github.com/tamland/python-tidal/compare/v0.6.7...v0.6.8) library changed over to using an OAuth flow. This PR updates mopidy-tidal to use the new flow. Note the changes in the README.

Curious what people think of this approach. Maybe as an alternative, it would be better to have a separate application that dumps out the OAuth tokens, and then you have to add them as part of the mopidy-tidal config?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,6 +49,15 @@ For High and Low quality be sure to have installed gstreamer bad-plugins, for eg
 
 This is mandatory to be able to play m4a streams.
 
+OAuth Flow
+----------
+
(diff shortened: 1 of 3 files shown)


tehkillerbee:
What you suggest is similar to PR #27 I think. But it requires a separate web page that can be opened on the mopidy server. This could be an alternative to having to to through the logs every time. It is however unfortunate, that the login has to be repeated for each restart. 

OAuth login should make it possible to use HI_RES quality in the future so it is nice that it is finally supported by tidalapi.

PR #36, opened 2021-04-17
Reload existing OAuth session from file. Updated version, readme

tehkillerbee:


greggilbert:
@tehkillerbee Good call, that's a much better way to do it. Did you bump the pypi version?

mopidy_tidal/backend.py
```diff
@@ -27,12 +30,45 @@ def __init__(self, config, audio):
...
+            with open(oauth_file, 'w') as outfile:
+                json.dump(data, outfile)
+
     def on_start(self):
         quality = self._config['tidal']['quality']
         logger.info("Connecting to TIDAL.. Quality = %s" % quality)
         config = Config(quality=Quality(quality))
         self._session = Session(config)
-        if self._session.login_oauth_simple(function=logger.info):
+        oauth_file = os.path.join(os.getenv("HOME"), '.config/', 'oauth.json')
```
greggilbert, line 51: Nitpicking maybe, but should this be `tidal-oauth.json` since it's in the shared `.config` folder?

tehkillerbee: Good point, I will update it accordingly. I have yet to update the pypi version, in case you want to test the latest version.

greggilbert: @tehkillerbee I pulled the latest and tested it, and it worked. 👍

PR #44, opened 2022-01-19
added client_id & client_secret to [tidal] in mopidy config

Glog78:
client_id & client_secret will be used instead of default
from tidalapi if defined in mopidy config

as requested pull request

### DIFF
--- a/mopidy_tidal/backend.py
+++ b/mopidy_tidal/backend.py
@@ -47,15 +47,20 @@ class TidalBackend(ThreadingActor, backend.Backend):
         config = Config(quality=Quality(quality))
         client_id = self._config['tidal']['client_id']
         client_secret = self._config['tidal']['client_secret']
-        if client_id and client_id.strip():
-            logger.info("Connecting to TIDAL.. Client ID read from config .. Client ID = %s" % client_id)
+
+        if (client_id and not client_secret) or (client_secret and not client_id):
+            logger.warn("Connecting to TIDAL.. always provide client_id and client_secret together")
+            logger.info("Connecting to TIDAL.. client id & client secret from pyhtontidal are used")
+        
+        if client_id and client_secret:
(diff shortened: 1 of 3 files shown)


mopidy_tidal/backend.py
```diff
@@ -45,6 +45,17 @@ def on_start(self):
...
+        client_secret = self._config['tidal']['client_secret']
+        if client_id and client_id.strip():
+            logger.info("Connecting to TIDAL.. Client ID read from config .. Client ID = %s" % client_id)
+            config.client_id=client_id
+            config.api_token=client_id
+            if client_secret and client_secret.strip():
+                logger.info("Connecting to TIDAL.. Client Secret read from config .. Client Secret = %s" % client_secret)
+                config.client_secret=client_secret
+            else:
+                logger.warn("Connecting to TIDAL.. client_id without client_secret will fail")
```
tehkillerbee, line 58: This warning will never be displayed, if neither client_id nor client_secret is set in config

Glog78: i checked it ... if you set only client_id but no client_secret this warning will stop the extension to work

Glog78: but let me see if i can redo this part

tehkillerbee: I misread your warning. I thought you wanted to have a warning if neither client_id/client_secret was set. 

But since you want to be able to fallback to using the built-in client_id/client_secret, I guess you can leave it as it is for now

tehkillerbee:
I've tested it and it works as expected with a working client_id / client_secret. I will need to update the README accordingly, since the user need to provide this key themselves.

Glog78:
;) last rework for better readable code and for better logging

tehkillerbee:
@Glog78 Thanks for the great work!

PR #45, opened 2022-01-19
Changed tidal-oauth output directory

tehkillerbee:
Determine the correct mopidy data_dir from the configuration file, instead of using HOME env. variable. This should make sure that the user has the correct permissions to write the tidal-oauth json file.

Updated setup.py to use latest tidalapi, as this version updates the API key.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,17 +32,23 @@ Configuration
 =============
 
 Before starting Mopidy, you must add configuration for
-Mopidy-Tidal to your Mopidy configuration file::
+Mopidy-Tidal to your Mopidy configuration file, if it is not already present.
+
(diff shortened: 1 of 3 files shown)


kingosticks:
[Extension.get_data_dir](https://docs.mopidy.com/en/latest/_modules/mopidy/ext/#Extension.get_data_dir) exists for this purpose. Something like:
```
from mopidy_tidal import Extension

oauth_file = Extension.get_data_dir(self.config) / "tidal-oauth.json"
```

tehkillerbee:
@kingosticks Thanks for the heads-up. Thats a cleaner solution. I will update it in the next version.

PR #60, opened 2022-05-22
Persist caches to disk to avoid slow initialization upon restart

BlackLight:
Note: this is still work in progress, but it already considerably reduces the extension bootstrap time upon restart.

Currently implemented:

- [x] Playlists cache
- [x] Tracks cache
- [x] Artists cache
- [x] Albums cache
- [x] Search cache
- [x] Unify the caches: they should all share the same `LruCache` base class, with filesystem caching support

### DIFF
--- a/mopidy_tidal/library.py
+++ b/mopidy_tidal/library.py
@@ -6,11 +6,15 @@ from requests.exceptions import HTTPError
 
 from mopidy import backend, models
 
-from mopidy.models import Image, SearchResult
+from mopidy.models import Image, Playlist, SearchResult
 
 from mopidy_tidal import full_models_mappers, ref_models_mappers
 
 from mopidy_tidal.lru_cache import LruCache
+
+from mopidy_tidal.helpers import to_timestamp
+
+from mopidy_tidal.playlists import PlaylistCache
 
 from mopidy_tidal.utils import apply_watermark
 
@@ -27,7 +31,9 @@ class TidalLibraryProvider(backend.LibraryProvider):
         self.lru_artist_img = LruCache(persist=False)
         self.lru_album_img = LruCache(persist=False)
         self.lru_playlist_img = LruCache(persist=False)
+        self.lru_artist_tracks = LruCache(persist=False)
         self.track_cache = LruCache()
+        self.playlist_cache = PlaylistCache()
 
     def get_distinct(self, field, query=None):
         from mopidy_tidal.search import tidal_search
@@ -141,6 +147,13 @@ class TidalLibraryProvider(backend.LibraryProvider):
             logger.info("EX")
             logger.info("%r", ex)
 
+    @staticmethod
+    def _get_image_uri(obj):
+        try:
+            return obj.image
+        except AttributeError:
+            pass
+
     def get_images(self, uris):
         logger.info("Searching Tidal for images for %r" % uris)
         session = self.backend._session
@@ -155,8 +168,9 @@ class TidalLibraryProvider(backend.LibraryProvider):
                         img_uri = self.lru_artist_img.get(artist_id)
                         if img_uri is None:
                             artist = session.get_artist(artist_id)
-                            img_uri = artist.image
-                            self.lru_artist_img[artist.id] = img_uri
+                            img_uri = self._get_image_uri(artist)
+                            if img_uri:
+                                self.lru_artist_img[artist.id] = img_uri
 
                         uri_images = [Image(uri=img_uri, width=512, height=512)]
                     elif parts[1] == 'album':
@@ -165,8 +179,9 @@ class TidalLibraryProvider(backend.LibraryProvider):
 
                         if img_uri is None:
                             album = session.get_album(album_id)
-                            img_uri = album.image
-                            self.lru_album_img[album_id] = img_uri
+                            img_uri = self._get_image_uri(album)
+                            if img_uri:
+                                self.lru_album_img[album_id] = img_uri
 
                         uri_images = [Image(uri=img_uri, width=512, height=512)]
                     elif parts[1] == 'playlist':
@@ -174,8 +189,9 @@ class TidalLibraryProvider(backend.LibraryProvider):
                         img_uri = self.lru_playlist_img.get(uri)
                         if img_uri is None:
                             playlist = session.get_playlist(playlist_id)
-                            img_uri = playlist.image
-                            self.lru_playlist_img[playlist_id] = img_uri
+                            img_uri = self._get_image_uri(playlist)
+                            if img_uri:
+                                self.lru_playlist_img[playlist_id] = img_uri
 
                         uri_images = [Image(uri=img_uri, width=512, height=512)]
                     elif parts[1] == 'track':
@@ -184,14 +200,15 @@ class TidalLibraryProvider(backend.LibraryProvider):
 
                         if img_uri is None:
                             album = session.get_album(album_id)
-                            img_uri = album.image
-                            self.lru_album_img[album_id] = img_uri
+                            img_uri = self._get_image_uri(album)
+                            if img_uri:
+                                self.lru_album_img[album_id] = img_uri
 
                         uri_images = [Image(uri=img_uri, width=512, height=512)]
 
                 images[uri] = uri_images or ()
-            except AttributeError:
-                logger.error("AttributeError when processing URI %r" % uri)
+            except AttributeError as err:
+                logger.error("AttributeError when processing URI %r: %s", uri, err)
             except HTTPError as err:
                 logger.error("HTTPError when processing URI %r: %s", uri, err)
 
@@ -209,21 +226,31 @@ class TidalLibraryProvider(backend.LibraryProvider):
         for uri in uris:
             try:
                 parts = uri.split(':')
-                # TODO General-purpose logic based on `tidal:<type>`
-                if uri.startswith('tidal:track:'):
-                    try:
-                        tracks.append(self.track_cache[uri])
-                    except KeyError:
-                        tracks += self._lookup_track(session, parts)
-                elif uri.startswith('tidal:album'):
-                    tracks += self._lookup_album(session, parts)
-                elif uri.startswith('tidal:artist'):
-                    tracks += self._lookup_artist(session, parts)
-                elif uri.startswith('tidal:playlist'):
-                    tracks += self._lookup_playlist(session, parts)
+                cache = None
+                lookup = None
 
-            except AttributeError:
-                logger.error("AttributeError when processing URI %r" % uri)
+                if uri.startswith('tidal:track:'):
+                    cache = self.track_cache
+                    lookup = self._lookup_track
+                elif uri.startswith('tidal:album'):
+                    cache = self.lru_album_tracks
+                    lookup = self._lookup_album
+                elif uri.startswith('tidal:artist'):
+                    cache = self.lru_artist_tracks
+                    lookup = self._lookup_artist
+                elif uri.startswith('tidal:playlist'):
+                    cache = self.playlist_cache
+                    lookup = self._lookup_playlist
+
+                if cache is not None and lookup:
+                    try:
+                        tracks.append(cache[uri])
+                    except KeyError:
+                        tracks += lookup(session, parts)
+                else:
+                    logger.debug('Unknown URI type: %s', uri)
+            except AttributeError as err:
+                logger.error("AttributeError when processing URI %r: %s" % uri, err)
             except HTTPError as err:
                 logger.error("HTTPError when processing URI %r: %s", uri, err)
 
@@ -232,12 +259,24 @@ class TidalLibraryProvider(backend.LibraryProvider):
         return tracks
 
     def _lookup_playlist(self, session, parts):
-        tracks = session.get_playlist_tracks(parts[2])
+        playlist_uri = ':'.join(parts)
+        tracks = self.playlist_cache.get(playlist_uri)
+
+        if tracks is None:
+            pl = session.get_playlist(playlist_uri)
+            tracks = session.get_playlist_tracks(parts[2])
+            self.playlist_cache[playlist_uri] = Playlist(
+                uri=playlist_uri,
+                name=pl.name,
+                tracks=tracks,
+                last_modified=to_timestamp(pl.last_updated),
+            )
+
         return full_models_mappers.create_mopidy_tracks(tracks)
 
     def _lookup_track(self, session, parts):
         album_id = parts[3]
-        album_uri = f'tidal:album:{parts[2]}:{album_id}'
+        album_uri = ':'.join(parts)
 
         tracks = self.lru_album_tracks.get(album_uri)
         if tracks is None:
@@ -251,10 +290,22 @@ class TidalLibraryProvider(backend.LibraryProvider):
 
     def _lookup_album(self, session, parts):
         album_id = parts[2]
-        tracks = session.get_album_tracks(album_id)
+        album_uri = ':'.join(parts)
+
+        tracks = self.lru_album_tracks.get(album_uri)
+        if tracks is None:
+            tracks = session.get_album_tracks(album_id)
+            self.lru_album_tracks[album_uri] = tracks
+
         return full_models_mappers.create_mopidy_tracks(tracks)
 
     def _lookup_artist(self, session, parts):
         artist_id = parts[2]
-        artist_tracks = session.get_artist_top_tracks(artist_id)
-        return full_models_mappers.create_mopidy_tracks(artist_tracks)
+        artist_uri = ':'.join(parts)
+
+        tracks = self.playlist_cache.get(artist_uri)
+        if tracks is None:
+            tracks = session.get_artist_top_tracks(artist_id)
+            self.lru_artist_tracks[artist_uri] = tracks
+
+        return full_models_mappers.create_mopidy_tracks(tracks)
--- a/mopidy_tidal/library.py
+++ b/mopidy_tidal/library.py
@@ -1,18 +1,20 @@
 from __future__ import unicode_literals
 
 import logging
+from typing import Optional, List
 
 from requests.exceptions import HTTPError
 
 from mopidy import backend, models
 
-from mopidy.models import Image, Playlist, SearchResult
+from mopidy.models import Image, SearchResult
 
-from mopidy_tidal import full_models_mappers, ref_models_mappers
+from mopidy_tidal import (
+    full_models_mappers,
+    ref_models_mappers,
+)
 
 from mopidy_tidal.lru_cache import LruCache
-
-from mopidy_tidal.helpers import to_timestamp
 
 from mopidy_tidal.playlists import PlaylistCache
 
@@ -27,19 +29,21 @@ class TidalLibraryProvider(backend.LibraryProvider):
 
     def __init__(self, *args, **kwargs):
         super(TidalLibraryProvider, self).__init__(*args, **kwargs)
-        self.lru_album_tracks = LruCache(max_size=10)
-        self.lru_artist_img = LruCache(persist=False)
-        self.lru_album_img = LruCache(persist=False)
-        self.lru_playlist_img = LruCache(persist=False)
-        self.lru_artist_tracks = LruCache(persist=False)
-        self.track_cache = LruCache()
-        self.playlist_cache = PlaylistCache()
+        self._artist_cache = LruCache()
+        self._album_cache = LruCache()
+        self._track_cache = LruCache()
+        self._playlist_cache = PlaylistCache()
+        self._image_cache = LruCache(directory='image')
+
+    @property
+    def _session(self):
+        return self.backend._session   # type: ignore
 
     def get_distinct(self, field, query=None):
         from mopidy_tidal.search import tidal_search
 
         logger.debug("Browsing distinct %s with query %r", field, query)
-        session = self.backend._session
+        session = self._session
 
         if not query:  # library root
             if field == "artist" or field == "albumartist":
@@ -76,7 +80,7 @@ class TidalLibraryProvider(backend.LibraryProvider):
         if not uri or not uri.startswith("tidal:"):
             return []
 
-        session = self.backend._session
+        session = self._session
 
         # summaries
 
@@ -137,7 +141,7 @@ class TidalLibraryProvider(backend.LibraryProvider):
 
         try:
             artists, albums, tracks = \
-                tidal_search(self.backend._session,
+                tidal_search(self._session,
                              query=query,
                              exact=exact)
             return SearchResult(artists=artists,
@@ -154,134 +158,102 @@ class TidalLibraryProvider(backend.LibraryProvider):
         except AttributeError:
             pass
 
+    def _get_image(self, uri) -> Optional[List[Image]]:
+        assert uri.startswith('tidal:'), f'Invalid TIDAL URI: {uri}'
+
+        parts = uri.split(':')
+        item_type = parts[1]
+        if item_type == 'track':
+            # For tracks, retrieve the artwork of the associated album
+            item_type = 'album'
+            item_id = parts[3]
+            uri = ':'.join([parts[0], 'album', parts[3]])
+        else:
+            item_id = parts[2]
+
+        if uri in self._image_cache:
+            # Cache hit
+            return self._image_cache[uri]
+
+        logger.debug('Retrieving %r from the API', uri)
+        getter_name = f'get_{item_type}'
+        getter = getattr(self._session, getter_name, None)
+        assert getter, f'No such session method: {getter_name}'
+
+        item = getter(item_id)
+        if not item:
+            logger.debug('%r is not available on the backend', uri)
+            return None
+
+        img_uri = self._get_image_uri(item)
+        if not img_uri:
+            return None
+
+        return [Image(uri=img_uri, width=320, height=320)]
+
     def get_images(self, uris):
         logger.info("Searching Tidal for images for %r" % uris)
-        session = self.backend._session
         images = {}
+
         for uri in uris:
             try:
-                uri_images = None
-                if uri.startswith('tidal:'):
-                    parts = uri.split(':')
-                    if parts[1] == 'artist':
-                        artist_id = parts[2]
-                        img_uri = self.lru_artist_img.get(artist_id)
-                        if img_uri is None:
-                            artist = session.get_artist(artist_id)
-                            img_uri = self._get_image_uri(artist)
-                            if img_uri:
-                                self.lru_artist_img[artist.id] = img_uri
+                images[uri] = self._get_image(uri)
+            except (AssertionError, AttributeError, HTTPError) as err:
+                logger.error(
+                    "%s when processing URI %r: %s",
+                    type(err), uri, err)
 
-                        uri_images = [Image(uri=img_uri, width=512, height=512)]
-                    elif parts[1] == 'album':
-                        album_id = parts[2]
-                        img_uri = self.lru_album_img.get(uri)
-
-                        if img_uri is None:
-                            album = session.get_album(album_id)
-                            img_uri = self._get_image_uri(album)
-                            if img_uri:
-                                self.lru_album_img[album_id] = img_uri
-
-                        uri_images = [Image(uri=img_uri, width=512, height=512)]
-                    elif parts[1] == 'playlist':
-                        playlist_id = parts[2]
-                        img_uri = self.lru_playlist_img.get(uri)
-                        if img_uri is None:
-                            playlist = session.get_playlist(playlist_id)
-                            img_uri = self._get_image_uri(playlist)
-                            if img_uri:
-                                self.lru_playlist_img[playlist_id] = img_uri
-
-                        uri_images = [Image(uri=img_uri, width=512, height=512)]
-                    elif parts[1] == 'track':
-                        album_id = parts[3]
-                        img_uri = self.lru_album_img.get(uri)
-
-                        if img_uri is None:
-                            album = session.get_album(album_id)
-                            img_uri = self._get_image_uri(album)
-                            if img_uri:
-                                self.lru_album_img[album_id] = img_uri
-
-                        uri_images = [Image(uri=img_uri, width=512, height=512)]
-
-                images[uri] = uri_images or ()
-            except AttributeError as err:
-                logger.error("AttributeError when processing URI %r: %s", uri, err)
-            except HTTPError as err:
-                logger.error("HTTPError when processing URI %r: %s", uri, err)
-
+        self._image_cache.update(images)
         return images
 
     def lookup(self, uris=None):
         logger.info("Lookup uris %r", uris)
-        session = self.backend._session
         if isinstance(uris, str):
             uris = [uris]
         if not hasattr(uris, '__iter__'):
             uris = [uris]
 
         tracks = []
-        for uri in uris:
+        for uri in (uris or []):
+            data = []
             try:
                 parts = uri.split(':')
-                cache = None
-                lookup = None
+                cache = getattr(self, f'_{parts[1]}_cache')
+                lookup = getattr(self, f'_lookup_{parts[1]}')
 
-                if uri.startswith('tidal:track:'):
-                    cache = self.track_cache
-                    lookup = self._lookup_track
-                elif uri.startswith('tidal:album'):
-                    cache = self.lru_album_tracks
-                    lookup = self._lookup_album
-                elif uri.startswith('tidal:artist'):
-                    cache = self.lru_artist_tracks
-                    lookup = self._lookup_artist
-                elif uri.startswith('tidal:playlist'):
-                    cache = self.playlist_cache
-                    lookup = self._lookup_playlist
+                try:
+                    data = cache[uri]
+                except KeyError:
+                    data = lookup(self._session, parts)
 
-                if cache is not None and lookup:
-                    try:
-                        tracks.append(cache[uri])
-                    except KeyError:
-                        tracks += lookup(session, parts)
-                else:
-                    logger.debug('Unknown URI type: %s', uri)
-            except AttributeError as err:
-                logger.error("AttributeError when processing URI %r: %s" % uri, err)
-            except HTTPError as err:
-                logger.error("HTTPError when processing URI %r: %s", uri, err)
+                tracks += data if hasattr(data, '__iter__') else [data]
+            except (AttributeError, HTTPError) as err:
+                logger.error("%s when processing URI %r: %s", type(err), uri, err)
 
         logger.info("Returning %d tracks", len(tracks))
-        self.track_cache.update({track.uri:track for track in tracks})
+        self._track_cache.update({track.uri:track for track in tracks})
         return tracks
 
     def _lookup_playlist(self, session, parts):
         playlist_uri = ':'.join(parts)
-        tracks = self.playlist_cache.get(playlist_uri)
+        playlist = self._playlist_cache.get(playlist_uri)
 
-        if tracks is None:
+        if playlist is None:
             pl = session.get_playlist(playlist_uri)
             tracks = session.get_playlist_tracks(parts[2])
-            self.playlist_cache[playlist_uri] = Playlist(
-                uri=playlist_uri,
-                name=pl.name,
-                tracks=tracks,
-                last_modified=to_timestamp(pl.last_updated),
+            playlist = self._playlist_cache[playlist_uri] = (
+                full_models_mappers.create_mopidy_playlist(pl, tracks)
             )
 
-        return full_models_mappers.create_mopidy_tracks(tracks)
+        return full_models_mappers.create_mopidy_tracks(playlist.tracks)
 
     def _lookup_track(self, session, parts):
         album_id = parts[3]
-        album_uri = ':'.join(parts)
+        album_uri = ':'.join(['tidal', 'album', album_id])
 
-        tracks = self.lru_album_tracks.get(album_uri)
+        tracks = self._album_cache.get(album_uri)
         if tracks is None:
             tracks = session.get_album_tracks(album_id)
-            self.lru_album_tracks[album_uri] = tracks
 
         track = [t for t in tracks if t.id == int(parts[4])][0]
         artist = full_models_mappers.create_mopidy_artist(track.artist)
@@ -292,20 +264,20 @@ class TidalLibraryProvider(backend.LibraryProvider):
         album_id = parts[2]
         album_uri = ':'.join(parts)
 
-        tracks = self.lru_album_tracks.get(album_uri)
+        tracks = self._album_cache.get(album_uri)
         if tracks is None:
             tracks = session.get_album_tracks(album_id)
-            self.lru_album_tracks[album_uri] = tracks
 
-        return full_models_mappers.create_mopidy_tracks(tracks)
+        self._album_cache[album_uri] = full_models_mappers.create_mopidy_tracks(tracks)
+        return self._album_cache[album_uri]
 
     def _lookup_artist(self, session, parts):
         artist_id = parts[2]
         artist_uri = ':'.join(parts)
 
-        tracks = self.playlist_cache.get(artist_uri)
+        tracks = self._artist_cache.get(artist_uri)
         if tracks is None:
             tracks = session.get_artist_top_tracks(artist_id)
-            self.lru_artist_tracks[artist_uri] = tracks
 
-        return full_models_mappers.create_mopidy_tracks(tracks)
+        self._artist_cache[artist_uri] = full_models_mappers.create_mopidy_tracks(tracks)
+        return self._artist_cache[artist_uri]
--- a/mopidy_tidal/library.py
+++ b/mopidy_tidal/library.py
@@ -187,8 +187,10 @@ class TidalLibraryProvider(backend.LibraryProvider):
 
         img_uri = self._get_image_uri(item)
         if not img_uri:
+            logger.debug('%r has no associated images', uri)
             return None
 
+        logger.debug('Image URL for %r: %r', uri, img_uri)
         return [Image(uri=img_uri, width=320, height=320)]
 
     def get_images(self, uris):
--- a/mopidy_tidal/library.py
+++ b/mopidy_tidal/library.py
@@ -29,11 +29,12 @@ logger = logging.getLogger(__name__)
 class TracksCache(OrderedDict):
     '''cache class, requires python3 style (ordered) dicts'''
 
-    def __init__(self, *args, backend, max_items=1000, **kwargs):
+    def __init__(self, *args, backend, max_items=1000, sync_fs_on_delete=False, **kwargs):
         super().__init__(*args, **kwargs)
         self.backend = backend
         self.max_items = max_items
         self.cache_dir = os.path.join(self.backend.cache_dir, 'tracks')
+        self._sync_fs_on_delete = sync_fs_on_delete
         pathlib.Path(self.cache_dir).mkdir(parents=True, exist_ok=True)
 
     def _cache_filename(self, key) -> str:
@@ -69,10 +70,11 @@ class TracksCache(OrderedDict):
 
     def __delitem__(self, key, *args, **kwargs):
         super().__delitem__(key, *args, **kwargs)
-        # Remove the entry also from the filesystem cache
-        cache_file = self._cache_filename(key)
-        if os.path.isfile(cache_file):
-            os.unlink(cache_file)
+        if self._sync_fs_on_delete:
+            # Remove the entry also from the filesystem cache
+            cache_file = self._cache_filename(key)
+            if os.path.isfile(cache_file):
+                os.unlink(cache_file)
 
     def update(self, d):
         super().update(d)
@@ -91,7 +93,7 @@ class TidalLibraryProvider(backend.LibraryProvider):
         self.lru_artist_img = LruCache()
         self.lru_album_img = LruCache()
         self.lru_playlist_img = LruCache()
-        self.track_cache = TracksCache(backend=self.backend)
+        self.track_cache = TracksCache(backend=self.backend, sync_fs_on_delete=False)
 
     def get_distinct(self, field, query=None):
         logger.debug("Browsing distinct %s with query %r", field, query)
--- a/mopidy_tidal/library.py
+++ b/mopidy_tidal/library.py
@@ -219,24 +219,33 @@ class TidalLibraryProvider(backend.LibraryProvider):
             uris = [uris]
 
         tracks = []
+        cache_updates = {}
+
         for uri in (uris or []):
             data = []
             try:
                 parts = uri.split(':')
-                cache = getattr(self, f'_{parts[1]}_cache')
+                cache_name = f'_{parts[1]}_cache'
+                cache = getattr(self, cache_name)
                 lookup = getattr(self, f'_lookup_{parts[1]}')
 
                 try:
                     data = cache[uri]
                 except KeyError:
-                    data = lookup(self._session, parts)
+                    data, parent = lookup(self._session, parts)
+                    if parent:
+                        cache_updates[cache_name] = cache_updates.get(cache_name, {})
+                        cache_updates[cache_name][uri] = parent
 
                 tracks += data if hasattr(data, '__iter__') else [data]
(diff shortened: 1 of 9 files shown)


tehkillerbee:
@BlackLight Great to see you've made progress! I agree, refactoring the caches used in Mopidy-Tidal is a good idea. I noticed the same thing back when I started maintaining this project but did not find the time to do it. (If it works, right?).
I am no expert on cache frameworks so I cant say for sure which type that would work best for all use-cases - or if the use of different cache types were intentional by the previous developer.

I'm looking forward to test it later today. We also need to make sure that the default directory is suitable as a cache directory. I'd look at Mopidy documentation for pointers on where these data should be stored: [link](https://docs.mopidy.com/en/latest/config/#core-section)

fmarzocca:
Other caches are here:` /var/lib/mopidy/.cache/`

BlackLight:
> Other caches are here:` /var/lib/mopidy/.cache/`

Good point. I'll prepare a commit to change the cache dir to `core/cache_dir` instead of creating a subfolder under `core/data_dir`.

BlackLight:
The persisted cache on storage has been implemented and the caches in the extension have been refactored. We now have a single `LruCache` class that implements (optional) filesystem persistence, and all the caches are now using it.

The LRU cache keeps a limited amount of items in memory (default: 1024) and all the other cached items are persisted on an indexed cache on the filesystem. The structure of the fs cache is inspired by that of `mopidy-spotify`:

```
<cache-dir>
  -> <type>
    -> <prefix>  # First two chars of the TIDAL ID
      -> tidal:<type>:<tidal-id>
```

In case of a miss on the memory cache, the filesystem cache will be inspected, and in case of a miss on the fs as well the extension would perform a lookup over API.

Some design decisions:

- The search cache is not persisted. Search results can change over time and I don't want to go down the rabbit hole of cache invalidation for search results.

- Image caches are not persisted. Caching them would blow the size of the cache directory. The price to pay is that images will be re-downloaded on-demand upon restart, but I think it's a quite reasonable trade-off.

Still pending:

- I need to properly test it with playlist updates. Things should work in theory (upon `load` we compare the stored `last_modified` of the cached Mopidy playlist with the `last_updated` of the TIDAL playlist), but I'd like to properly validate it.

- I need to test artist caches.

Testing volunteers are welcome!

tehkillerbee:
@BlackLight Great work, looking forward to test it!

adamcik:
I don't disagree that this type of cache case is probably shared by multiple plugins, but I'm not sure if adding this to mopidy itself really makes sense. In a similar vein I've been wanting to factor out a bunch of the webservice/oauth type helpers so that we can just reuse things more. But I suspect that it might be better to just have someone create a `mopidy-client-libs` or a `mopidy-plugin-helpers` (or whatever a suitable name is).

If there are parts of this that can only be solved in mopidy, and there is real value to having it solved in that codebase then definitely open a bug with why this is the case. Otherwise I think this would benefit more from not being in "core", similar to how we've moved e.g. `mopidy-mpd` out so it can be improved independently of mopidy releases.

BlackLight:
@adamcik I totally agree - if the plan is to keep the core "lightweight" then non-core features should be moved out of the core, not in.

There's still a valid point though for features that are non-core but still shared by many extensions (such as caching, OAuth flows and playlist updates). Caching and OAuth in particular are implemented independently by multiple extensions (Spotify, SoundCloud, Tidal, YouTube etc.) even though they mostly share the same code - and, in some cases, most of the code of these features is copy-pasted from other extensions.

I like the idea of something like a `mopidy-ext-helpers` package that could collect most of these features, as well as provide some room for customization/extensibility on top of the boilerplate. We would have a long tail of extensions that should probably migrate at some point, but it's good to at least lay out the foundations.

BlackLight:
The PR is ready for test/review.

The latest changes involve a more consistent cache implementation/usage in `library.lookup`. I have tested the caching features for artists and albums across restarts with Iris and they also seem to work fine.

I have also fixed the logic for playlist cache invalidation upon updates. We have a problem because the current stable branch of `python-tidal` doesn't currently expose the `last_updated` attribute of the Tidal playlists. It's [apparently implemented on the `0.7.x` branch](https://github.com/tamland/python-tidal/blob/0.7.x/tidalapi/playlist.py#L74), but that version isn't on pip yet (the author said that it should be coming soon, but after waiting about two years I'm not sure of their definition of "soon"). In the meantime, [I have pushed a fix myself](https://github.com/tamland/python-tidal/pull/99/commits/7490fc4a1ebd6f97c750d30f2262e12c81e4698a) in my existing PR for pagination, so you may want to use the `pagination-support` branch from my fork of `python-tidal` if you want both pagination, and your playlists to be up-to-date when you add/remove tracks.

tehkillerbee:
Browsing appears very fast indeed! Only issue I see is the images are continuously requested from the server - not from the cache. The issue is easy to replicate. Wait for the images to be populated, navigate into a specific playlist/artist folder and then navigate back to the root folder. Images will then be requested again.

tehkillerbee:
Sometimes when refreshing a directory, an error is thrown:
```
May 26 10:44:15 jlinde-ThinkPad-T430 mopidy[8625]: ERROR    [TidalBackend-6] mopidy_tidal.library HTTPError when processing URI 'tidal:album:38264741': 404 Client Error: Not Found for url: https://api.tidalhifi.com/v1/albums/38264741?sessionId=<session>8&countryCode=DK&limit=999
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]: ERROR    [Core-7] mopidy.core.library TidalBackend backend caused an exception.
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]: Traceback (most recent call last):
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/lib/python3/dist-packages/mopidy/core/library.py", line 17, in _backend_error_handling
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     yield
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/lib/python3/dist-packages/mopidy/core/library.py", line 181, in get_images
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     if future.get() is None:
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/lib/python3/dist-packages/pykka/_threading.py", line 45, in get
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     _compat.reraise(*self._data['exc_info'])
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/lib/python3/dist-packages/pykka/_compat/__init__.py", line 29, in reraise
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     raise value
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/lib/python3/dist-packages/pykka/_actor.py", line 193, in _actor_loop
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     response = self._handle_receive(envelope.message)
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/lib/python3/dist-packages/pykka/_actor.py", line 299, in _handle_receive
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     return callee(*message.args, **message.kwargs)
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/local/lib/python3.8/dist-packages/Mopidy_Tidal-0.2.8-py3.8.egg/mopidy_tidal/library.py", line 181, in get_images
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     album = session.get_album(album_id)
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/local/lib/python3.8/dist-packages/tidalapi/__init__.py", line 385, in get_album
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     return self._map_request('albums/%s' % album_id, ret='album')
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/local/lib/python3.8/dist-packages/tidalapi/__init__.py", line 455, in _map_request
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     json_obj = self.request('GET', url, params).json()
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/local/lib/python3.8/dist-packages/tidalapi/__init__.py", line 357, in request
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     request = self.basic_request(method, path, params, data, headers)
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/local/lib/python3.8/dist-packages/tidalapi/__init__.py", line 348, in basic_request
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     if not request.ok and request.json()['userMessage'].startswith("The token has expired.") and self.refresh_token:
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/lib/python3/dist-packages/requests/models.py", line 897, in json
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     return complexjson.loads(self.text, **kwargs)
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/lib/python3/dist-packages/simplejson/__init__.py", line 518, in loads
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     return _default_decoder.decode(s)
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/lib/python3/dist-packages/simplejson/decoder.py", line 370, in decode
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     obj, end = self.raw_decode(s)
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:   File "/usr/lib/python3/dist-packages/simplejson/decoder.py", line 400, in raw_decode
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]:     return self.scan_once(s, idx=_w(s, idx).end())
May 26 10:44:20 jlinde-ThinkPad-T430 mopidy[8625]: simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

tehkillerbee:
When 404 occurs for a specific URI, the URI is not added to the cache.

Instead, we should probably add a dummy entry to the cache to avoid retrying on next refresh.

```
May 26 11:01:25 jlinde-ThinkPad-T430 mopidy[9505]: ERROR    [TidalBackend-6] mopidy_tidal.library HTTPError when processing URI 'tidal:album:38264741': 404 Client Error: Not Found for url: https://api.tidalhifi.com/v1/albums/38264741?sessionId=<session ID>&countryCode=DK&limit=999
```

mopidy_tidal/library.py
```diff
@@ -168,47 +165,52 @@ def get_images(self, uris):
...
-                            self.lru_artist_img[artist.id] = img_uri
+                            img_uri = self._get_image_uri(artist)
+                            if img_uri:
+                                self.lru_artist_img[artist.id] = img_uri
 
                         uri_images = [Image(uri=img_uri, width=512, height=512)]
                     elif parts[1] == 'album':
                         album_id = parts[2]
-                        img_uri = self.lru_album_img.hit(album_id)
+                        img_uri = self.lru_album_img.get(uri)
```
tehkillerbee, line 178: Shouldn't it be get(album_id) instead?

mopidy_tidal/library.py
```diff
@@ -168,47 +165,52 @@ def get_images(self, uris):
...
-                            self.lru_album_img[album_id] = img_uri
+                            img_uri = self._get_image_uri(album)
+                            if img_uri:
+                                self.lru_album_img[album_id] = img_uri
 
                         uri_images = [Image(uri=img_uri, width=512, height=512)]
                     elif parts[1] == 'playlist':
                         playlist_id = parts[2]
-                        img_uri = self.lru_playlist_img.hit(playlist_id)
+                        img_uri = self.lru_playlist_img.get(uri)
```
tehkillerbee, line 189: Shouldn't it be get(playlist_id) instead?

mopidy_tidal/library.py
```diff
@@ -168,47 +165,52 @@ def get_images(self, uris):
...
-                            self.lru_playlist_img[playlist_id] = img_uri
+                            img_uri = self._get_image_uri(playlist)
+                            if img_uri:
+                                self.lru_playlist_img[playlist_id] = img_uri
 
                         uri_images = [Image(uri=img_uri, width=512, height=512)]
                     elif parts[1] == 'track':
                         album_id = parts[3]
-                        img_uri = self.lru_album_img.hit(album_id)
+                        img_uri = self.lru_album_img.get(uri)
```
tehkillerbee, line 199: Shouldn't it be get(album_id) instead?

mopidy_tidal/library.py
```diff
@@ -157,6 +147,13 @@ def search(self, query=None, uris=None, exact=False):
             logger.info("EX")
             logger.info("%r", ex)
 
+    @staticmethod
+    def _get_image_uri(obj):
+        try:
+            return obj.image
```
tehkillerbee, line 153: Consider changing this to the below, so we can override the max resolution. Otherwise we will return really large images.
```
@staticmethod
def _get_image_uri(obj):
    try:
        return obj.picture(width=320, height=320)
    except AttributeError:
        pass
```

tehkillerbee (changes requested):

mopidy_tidal/library.py
```diff
@@ -168,47 +165,52 @@ def get_images(self, uris):
...
-                            img_uri = album.image
-                            self.lru_album_img[album_id] = img_uri
+                            img_uri = self._get_image_uri(album)
+                            if img_uri:
+                                self.lru_album_img[album_id] = img_uri
 
                         uri_images = [Image(uri=img_uri, width=512, height=512)]
-                        pass
 
                 images[uri] = uri_images or ()
```
tehkillerbee, line 209: Consider adding "logger.debug("img_uri %s", img_uri)"

mopidy_tidal/library.py
```diff
@@ -24,9 +24,9 @@ class TidalLibraryProvider(backend.LibraryProvider):
     def __init__(self, *args, **kwargs):
         super(TidalLibraryProvider, self).__init__(*args, **kwargs)
         self.lru_album_tracks = LruCache(max_size=10)
```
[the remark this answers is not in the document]
BlackLight: `max_size` defaults to 1024 and it only applies to the cache in memory (the cache on disk doesn't have limits). The `max_size=10` limit for the album tracks is probably a bit too conservative, we could use 1024 here.

kingosticks:
Am I right in saying that for every cache miss in `lookup()` you will write once to disk? Can you write them once in bulk at the end?

mopidy_tidal/library.py
BlackLight: See [here](https://github.com/tehkillerbee/mopidy-tidal/pull/60#discussion_r882739323)

mopidy_tidal/library.py
BlackLight: See [here](https://github.com/tehkillerbee/mopidy-tidal/pull/60#discussion_r882739323)

mopidy_tidal/library.py
BlackLight: Good one - I think it makes sense to limit the image resolution.

mopidy_tidal/library.py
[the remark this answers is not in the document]
tehkillerbee: @BlackLight agreed.

BlackLight:
> Perhaps we should consider a local image cache as well?

I thought of it, but I have kept things like this for now because I'm not sure about the storage requirements. AFAIK `mopidy-spotify` caches images too, and for this reason `mopidy-spotify`'s cache directory can easily get to a few GB in size. If a 320x320 image is about 40 KB in size then storing 10,000 of them (that's quite realistic after a few months/years spent browsing music) can easily result in ~400 MB of cache storage. Maybe there could be a configuration flag for caching images? It can be enabled by default, and if one is running the app on a system with low storage it can be disabled. And maybe it's also good to implement an LRU mechanism for this cache so it doesn't explode in size. @kingosticks any suggestions on this?

> Sometimes when refreshing a directory, an error is thrown. Perhaps this is just because of excessive Tidal requests?

I've also got the same error occasionally. Maybe it's worth investigating if it reproduces with the same track? The 404 error makes me think that it may be a temporarily unavailable track, or maybe a track not available in the user's market. If they're using a 404 for a limit exceeded error then they're doing HTTP wrong :)

> Instead, we should probably add a dummy entry to the cache to avoid retrying on next refresh.

It depends on what the error is IMHO. If 404 means "this track is not available (at all)", then it makes sense to add a dummy entry to avoid reloading it. If instead it's due to a service temporarily unavailable or a limit exceeded, then we want to try and reload it on the next lookup - and a dummy entry wouldn't allow us to refresh the cache entry. In my tests I have noticed that when these errors happen it's only the invalid track that gets a cache miss, not the whole playlists, so hopefully this shouldn't have a big performance impact.

> Am I right in saying that for every cache miss in lookup() you will write once to disk? Can you write them once in bulk at the end?

We cache tracks outside of the `for` in [`library.lookup`](https://github.com/tehkillerbee/mopidy-tidal/blob/831ebc09edb924c3aa98ddd35aa30093c7d503f9/mopidy_tidal/library.py#L258), but playlists, albums and tracks are indeed cached within the `for`. They shouldn't be as write-intensive as tracks, but it's indeed a good idea to write them to the cache outside of the cycle.

tehkillerbee:
> I thought of it, but I have kept things like this for now because I'm not sure about the storage requirements. AFAIK `mopidy-spotify` caches images too, and for this reason `mopidy-spotify`'s cache directory can easily get to a few GB in size. If a 320x320 image is about 40 KB in size then storing 10,000 of them (that's quite realistic after a few months/years spent browsing music) can easily result in ~400 MB of cache storage. Maybe there could be a configuration flag for caching images? It can be enabled by default, and if one is running the app on a system with low storage it can be disabled. And maybe it's also good to implement an LRU mechanism for this cache so it doesn't explode in size. @kingosticks any suggestions on this?

@BlackLight Interesting, I did not know mopidy-spotify already did this. My suggestion would also be to make the image caching an option that can be enabled if necessary. It would speed up the browsing experience when using clients that rely heavily on album art (such as Iris), that's for sure. 

> I've also got the same error occasionally. Maybe it's worth investigating if it reproduces with the same track? The 404 error makes me think that it may be a temporarily unavailable track, or maybe a track not available in the user's market. If they're using a 404 for a limit exceeded error then they're doing HTTP wrong :)

Yes, I see the 404 error every time I load my list of albums. Some of the albums I have added has probably been removed again by the artist. Unfortunately, tidalapi does not handle unavailable albums very well.

The following JSONDecodeError error is something I have not seen before the caching changes has been introduced. It seems to occur after refreshing the folder a few times within a short period of time. Perhaps it is just Tidal's response to too many repeated requests?

> It depends on what the error is IMHO. If 404 means "this track is not available (at all)", then it makes sense to add a dummy entry to avoid reloading it. If instead it's due to a service temporarily unavailable or a limit exceeded, then we want to try and reload it on the next lookup - and a dummy entry wouldn't allow us to refresh the cache entry. In my tests I have noticed that when these errors happen it's only the invalid track that gets a cache miss, not the whole playlists, so hopefully this shouldn't have a big performance impact.

Agreed, although I have only seen 404 errors on unavailable albums not as a general error. In any case, the albums are just left unpopulated in the list and the remaining albums can still be listed.

BlackLight:
> It seems to occur after refreshing the folder a few times within a short period of time.

It sounds like caching implies faster throughput -> more requests per second -> we're more likely to hit the API's throttling thresholds. But again, if that's indeed the case, they should probably return `429` (even a `403` would be better), not `404`. And, also, a JSON response even in case of error 👇

> The following JSONDecodeError error is something I have not seen before the caching changes has been introduced.

If the actual response code is 404 then it's likely that they just return HTML/text, hence the JSONDecodeError, but it may be worth to add a print to check that it's indeed the case.

mopidy_tidal/library.py
tehkillerbee: @BlackLight I do not think this works as intended. I have added some additional log messages to verify. Cache miss always occurs, even after first initialization. Try to add the following log output and you'll see what I mean:

```
...
elif parts[1] == 'album':
album_id = parts[2]
img_uri = self.lru_album_img.get(uri)

if img_uri is None:
>>    logger.info("adding %s to cache %s", parts[1], parts[2])
    album = session.get_album(album_id)
    img_uri = self._get_image_uri(album)
    if img_uri:
        self.lru_album_img[album_id] = img_uri
...
```

BlackLight:
@tehkillerbee indeed, it seems that we always have a cache miss on images. I didn't wire everything properly with the existing caching logic, and now I'm just in the middle of refactoring it so some of the repeated that can be easily parametrized goes away.

On a closer look of the existing logic (as well as the image lookup logic in [`mopidy-spotify`](https://github.com/mopidy/mopidy-spotify/blob/master/mopidy_spotify/images.py#L104)) it seems though that we always cache `Image` objects - I'm not sure if those eventually get cached like full images on `mopidy-spotify` or just URIs. I'll stick to the same logic for this PR - and if caching an `Image` object involves only caching the URL then an additional `cache_images` flag may be redundant.

kingosticks:
I'm a bit confused. As I remember, Mopidy-Spotify doesn't cache actual image data, just the result of the Spotify API call which has the actual image URI. Are you trying to cache the actual image data here? How are clients then getting access to that data? Do I misunderstand? I'm sorry but I've not reviewed the latest code fully.

BlackLight:
@kingosticks

> As I remember, Mopidy-Spotify doesn't cache actual image data, just the result of the Spotify API call which has the actual image URI.

You're right indeed. Initially I though that the `Image` object would perform a lazy inizialitation of the image data as well, but at a closer look I noticed that it only contains the data necessary to construct the image URL.

So to answer @tehkillerbee - I guess that we can't really talk of server-side full image caching: the web client is already in charge of caching static images usually, on the server we'll only cache the URLs.

I've also pushed new changes that involve a major rewrite of the caching logic in the `library` module (it was very spaghetti-like and full of copy-pasted section). This should also fix the issue with image cache misses, as well as rationalize the caches used in the code (we now use both the same lookup logic and the same cache objects, no more nested if's per URI type, no more different caches for track lists and object models). I've tested it both against Iris and mopidy for several cases (artists, albums, playlists and single tracks) and things seem to work.

fmarzocca:
IMHO there is no sense to cache the image data, if ever the images url...

tehkillerbee:
> IMHO there is no sense to cache the image data, if ever the images url...

@fmarzocca It appears to be faster to use the cached image URL instead of requesting it anew. So it will provide a better user experience and images will be displayed faster.

tehkillerbee:
@BlackLight These are my comments after testing. It appears there are a few things that appear broken (In Iris), mostly related to unavailable images. 

**1a**. Artist images are never loaded when opening "Artists" directory (no caching occurs). When clicking directory "My Artist", Images are requested but never displayed. Instead an error is shown.
```
May 29 20:32:32 jlinde-ThinkPad-T430 mopidy[110780]: INFO     [TidalBackend-6] mopidy_tidal.library Browsing uri tidal:my_artists
May 29 20:32:32 jlinde-ThinkPad-T430 mopidy[110780]: INFO     [TidalBackend-6] mopidy_tidal.library Searching Tidal for images for ['tidal:artist:3567061', 
...
'tidal:artist:9810587', 'tidal:artist:15295', 'tidal:artist:4130', 'tidal:artist:3588012', 'tidal:artist:8875626', 'tidal:artist:13686']
May 29 20:32:33 jlinde-ThinkPad-T430 mopidy[110780]: ERROR    [Core-7] mopidy.core.library TidalBackend backend returned bad data: Expected a list of Image, not None
```
**1b**. Playlists images are never loaded when opening "Playlists" directory (no caching occurs). Opening "My Playlists" directory results in partial loading of album art. Only some of the first playlists has an image. The remaining ones are blank.
```
May 29 20:26:09 jlinde-ThinkPad-T430 mopidy[108654]: INFO     [TidalBackend-6] mopidy_tidal.library Browsing uri tidal:my_playlists
May 29 20:26:09 jlinde-ThinkPad-T430 mopidy[108654]: INFO     [TidalBackend-6] mopidy_tidal.library Searching Tidal for images for ['tidal:playlist:ae24e029-6024-4fc4-9c83-f77487a77bc7', 'tidal:playlist:9d80a1e2-2f5d-4825-af0e-aa7ab4dd1ea9', 'tidal:playlist:eeade4f0-7874-4867-9346-5f35bb281d18', 'tidal:playlist:ff963aeb-2e4b-4a6b-a5e2-8da10740902b', 'tidal:playlist:58b26365-afc9-413c-9437-04b58ec9748f', 'tidal:playlist:cd1d80e2-43c3-45c8-8f10-9a8c91b7777e', 'tidal:playlist:9698473c-af02-42ec-b6fb-e9da82565cf4', 'tidal:playlist:b4d9fc6f-36ee-408c-ab92-94012af35d84', 'tidal:playlist:7ab5d2b6-93fb-4181-a008-a1d18e2cebfa', 'tidal:playlist:48085a2b-62cb-42ba-9e03-cdb7a9a9d9ab', 'tidal:playlist:1a7d78db-6f0f-4bbd-8223-bdd8826f7d7c', 'tidal:playlist:8b5fa9ad-7310-48c7-b468-0f574390e39f', 'tidal:playlist:76083a45-f5fa-4723-953c-09674ce48fdf', 'tidal:playlist:f7841e40-d38a-4de2-90f6-2326bf5f164a', 'tidal:playlist:c38b1def-0944-442a-94cf-b7d86e8c8ab0', 'tidal:playlist:f31bc2f9-118e-4359-bf74-16593a6aabc8', 'tidal:playlist:de60da1f-8269-419b-a9fc-9a9f15dfecab', 'tidal:playlist:11fa1642-2394-47d5-ad3a-4da653a4907e', 'tidal:playlist:f4400845-fd1c-434d-8da1-1e818124a417', 'tidal:playlist:17df59b6-fb51-4aa9-a1ab-f43386f20859', 'tidal:playlist:f974f0a9-026a-4e6a-b061-f458ab4bc634', 'tidal:playlist:2c30f6c0-d3ce-458b-9a0d-d0569d115014', 'tidal:playlist:578a0303-c876-4f1b-afd7-eb1e7f4e12fb']
May 29 20:26:10 jlinde-ThinkPad-T430 mopidy[108654]: ERROR    [Core-7] mopidy.core.library TidalBackend backend returned bad data: Expected a list of Image, not None
```
This error appears to occur when a user created playlist has no image. In general, most of the above are related to missing images not handled gracefully.

The issues with images not being requested when clicking "Album" and "Artist" directory on the left side of the web interface is probably an issue with Iris, as it also occurred before the changes.

**2.** Attempting to open "Playlists" (not My Playlists) directory, results in the following (VERY long log message), although playlists get displayed. Error is only shown immediately after clicking "refresh". 
```
May 29 20:20:54 jlinde-ThinkPad-T430 mopidy[108654]: INFO     [TidalBackend-6] mopidy_tidal.library Lookup uris 'tidal:playlist:17df59b6-fb51-4aa9-a1ab-f43386f20859'
May 29 20:20:54 jlinde-ThinkPad-T430 mopidy[108654]: INFO     [TidalBackend-6] mopidy_tidal.library Returning 1 tracks
May 29 20:20:54 jlinde-ThinkPad-T430 mopidy[108654]: ERROR    [Core-7] mopidy.core.library TidalBackend backend returned bad data: Expected a list of Track, not [Playlist(last_modified=0, name='Series theme', tracks=[Track(album=Album(artists=[Artist(name='Game of Thrones Orchestra', uri='tidal:artist:4090193')], name='Game of Thrones (Main Opening Theme from The Series)', uri='tidal:album:46026085'), artists=[Artist(name='Game of Thrones Orchestra', uri='tidal:artist:4090193')], disc_no=1, length=105000, name='Game of Thrones (Main Theme of the TV Series)', track_no=1, uri='tidal:track:4090193:46026085:46026086'), ...
```
After selecting one of the playlists, everything works fine and images are displayed

**3** It appears we should return a "dummy" image and not a "None" type to mopidy, otherwise the above errors are shown whenever no image exists for a given track/playlist/album.

**4** Currently, we have not set the max resolution to use when requesting images. We should consider changing it, as mentioned above. Otherwise we request images/album art that is 1080x1080. It looks like iris does not like images that are too large(?).

With all those bugs and comments out of the way, in general, the browsing experience is pretty good - after the first initial caching sequence.  Great work and very nice to see the spaghetti code refactored! We should make sure to update the changelog accordingly (and give credit!). PR's like this are greatly(!) appreciated.

BlackLight:
@tehkillerbee can you try and remove your cache directory and try again? In the previous implementation some things were cached as mopidy objects and some (the `lru_*_tracks` objects) just as lists of tracks.

I've made things more consistent by always caching things as mopidy objects (after all, that's what we usually want to unpack anyway), but this means that cached items created previously break the current implementation.

tehkillerbee:
> @tehkillerbee can you try and remove your cache directory and try again? In the previous implementation some things were cached as mopidy objects and some (the `lru_*_tracks` objects) just as lists of tracks.
> 
> I've made things more consistent by always caching things as mopidy objects (after all, that's what we usually want to unpack anyway), but this means that cached items created previously break the current implementation.

It looks like the caching works as expected. However certain folders (My Artists) still return an error and no album art. My Playlists only displays album art on some playlists. An error "TidalLibraryProvider object has no attribute xx_cache" is still thrown when browsing the various directories (My Artists, My Albums, My Tracks). 

* My Artists
```
May 30 17:35:46 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.library Browsing uri tidal:directory
May 30 17:35:48 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.library Lookup uris 'tidal:my_artists'
May 30 17:35:48 jlinde-ThinkPad-T430 mopidy[2551]: ERROR    [TidalBackend-6] mopidy_tidal.library <class 'AttributeError'> when processing URI 'tidal:my_artists': 'TidalLibraryProvider' object has no attribute '_my_artists_cache'
May 30 17:35:48 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.library Returning 0 tracks
May 30 17:35:48 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.library Browsing uri tidal:my_artists
May 30 17:35:48 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.library Searching Tidal for images for ['tidal:artist:3567061', 
...
'tidal:artist:9810587', 'tidal:artist:15295', 'tidal:artist:4130', 'tidal:artist:3588012', 'tidal:artist:8875626', 'tidal:artist:13686']
May 30 17:35:51 jlinde-ThinkPad-T430 mopidy[2551]: ERROR    [Core-7] mopidy.core.library TidalBackend backend returned bad data: Expected a list of Image, not None
```

* My Playlists
```
May 30 17:35:02 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.library Lookup uris 'tidal:my_playlists'
May 30 17:35:02 jlinde-ThinkPad-T430 mopidy[2551]: ERROR    [TidalBackend-6] mopidy_tidal.library <class 'AttributeError'> when processing URI 'tidal:my_playlists': 'TidalLibraryProvider' object has no attribute '_my_playlists_cache'
May 30 17:35:02 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.library Returning 0 tracks
May 30 17:35:02 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.library Browsing uri tidal:my_playlists
May 30 17:35:03 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.library Searching Tidal for images for ['tidal:playlist:ae24e029-6024-4fc4-9c83-f77487a77bc7', 'tidal:playlist:9d80a1e2-2f5d-4825-af0e-aa7ab4dd1ea9', 'tidal:playlist:eeade4f0-7874-4867-9346-5f35bb281d18', 'tidal:playlist:ff963aeb-2e4b-4a6b-a5e2-8da10740902b', 
...
'tidal:playlist:f974f0a9-026a-4e6a-b061-f458ab4bc634', 'tidal:playlist:2c30f6c0-d3ce-458b-9a0d-d0569d115014', 'tidal:playlist:578a0303-c876-4f1b-afd7-eb1e7f4e12fb']
May 30 17:35:03 jlinde-ThinkPad-T430 mopidy[2551]: ERROR    [Core-7] mopidy.core.library TidalBackend backend returned bad data: Expected a list of Image, not None
```

Manually refreshing playlists still results in a massive error log message:
```
May 30 17:38:32 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.playlists Refreshing TIDAL playlists..
May 30 17:38:43 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.playlists TIDAL playlists refreshed
May 30 17:38:43 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.library Browsing uri tidal:my_playlists
May 30 17:38:44 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.library Lookup uris 'tidal:playlist:ae24e029-6024-4fc4-9c83-f77487a77bc7'
May 30 17:38:44 jlinde-ThinkPad-T430 mopidy[2551]: INFO     [TidalBackend-6] mopidy_tidal.library Returning 1 tracks
May 30 17:38:44 jlinde-ThinkPad-T430 mopidy[2551]: ERROR    [Core-7] mopidy.core.library TidalBackend backend returned bad data: Expected a list of Track, not [Playlist(last_modified=0
...
```
Browsing the playlists themselves reveal playlists has 0 tracks in them. Manually refreshing the playlist works fine and all tracks are now shown.

Generally, it looks like we are not returning result data corresponding to the data expected by Mopidy. It seems the refactoring of lookup() function in library broke something, as we can now return a non-track list type to Mopidy.

mopidy_tidal/library.py
```diff
@@ -157,104 +151,111 @@ def search(self, query=None, uris=None, exact=False):
...
         if not hasattr(uris, '__iter__'):
             uris = [uris]
 
         tracks = []
-        for uri in uris:
+        for uri in (uris or []):
+            data = []
             try:
                 parts = uri.split(':')
-                if uri.startswith('tidal:track:'):
```
tehkillerbee, line 227: I understand the logic behind defining cache, lookup functions depending on the URI type but I think it is less readable that way. I prefer an if/else statement, similar to what we used previously:
```python
for uri in (uris or []):
data = []
try:
parts = uri.split(':')
if uri.startswith('tidal:track:'):
    cache = self._track_cache
    lookup = self._lookup_track
elif uri.startswith('tidal:album'):
    cache = self._album_cache
    lookup = self._lookup_album
elif uri.startswith('tidal:artist'):
    cache = self._track_cache
    lookup = self._lookup_artist
elif uri.startswith('tidal:playlist'):
    cache = self._playlist_cache
    lookup = self._lookup_playlist

try:
    data = cache[uri]
except KeyError:
    data = lookup(self._session, parts)
```

tehkillerbee (changes requested):

mopidy_tidal/library.py
```diff
@@ -157,104 +151,111 @@ def search(self, query=None, uris=None, exact=False):
...
-                logger.error("HTTPError when processing URI %r" % uri)
+                cache = getattr(self, f'_{parts[1]}_cache')
+                lookup = getattr(self, f'_lookup_{parts[1]}')
+
+                try:
+                    data = cache[uri]
+                except KeyError:
+                    data = lookup(self._session, parts)
+
+                tracks += data if hasattr(data, '__iter__') else [data]
```
tehkillerbee, line 231: I think this line should be `tracks += data if hasattr(data, '__iter__') else data.tracks` to fix the error `Expected a list of Track, not [Playlist..` when refreshing playlists

tehkillerbee (changes requested):

mopidy_tidal/library.py
```diff
@@ -157,104 +151,111 @@ def search(self, query=None, uris=None, exact=False):
...
-
-                        uri_images = [Image(uri=img_uri, width=512, height=512)]
-                        pass
-
-                images[uri] = uri_images or ()
-            except AttributeError:
-                logger.error("AttributeError when processing URI %r" % uri)
-            except HTTPError as err:
-                logger.error("HTTPError when processing URI %r" % uri)
+                images[uri] = self._get_image(uri)
```
tehkillerbee, line 202: Change this line to `images[uri] = self._get_image(uri) or ()` to fix the error `TidalBackend backend returned bad data: Expected a list of Image, not None` when browsing My Playlists, My Albums etc.

tehkillerbee (changes requested):

mopidy_tidal/library.py
tehkillerbee: It seems some URI types fall outside the above categories. This will cause the subsequent cache lookup to fail. For now, lets stick to your refactored version :)

mopidy_tidal/library.py
tehkillerbee: @BlackLight Perhaps we should reconsider changing the resolution, as the images are almost too small when browsing Iris. Clicking on the larger preview results in a tiny image.

It looks like all types support 750x750 (I did some tests with the tidal web player and images using this resolution exists for playlists, albums and artists, even though this is not mentioned in the tidalapi source comments).

tehkillerbee (changes requested):

fmarzocca:
> * Error is thrown whenever playlist directory is refreshed from Iris.

Not completely true. It is thrown whenever Mopidy receive a `[core.playlists.refresh](https://docs.mopidy.com/en/latest/api/core/#mopidy.core.PlaylistsController.refresh)` call.

mopidy_tidal/library.py
BlackLight: @tehkillerbee I have added 750x750 as a standard resolution. 512x512 would have probably been a better trade-off, but in some of my tests several items didn't have 512x512 images associated.

BlackLight:
> Am I right in saying that for every cache miss in `lookup()` you will write once to disk? Can you write them once in bulk at the end?

@kingosticks this is now addressed in https://github.com/tehkillerbee/mopidy-tidal/pull/60/commits/6416ed3c18355daa11c35b7833d35101a38ae416

mopidy_tidal/library.py
BlackLight: I have fixed the issue by letting `_lookup_playlist` return a `tracks, playlist` tuple instead of a list of tracks. For lookup purposes we need the list, but for caching purposes we need the playlist object.

mopidy_tidal/library.py
BlackLight: @tehkillerbee this should be fixed in https://github.com/tehkillerbee/mopidy-tidal/pull/60/commits/b92064caa62d4980185ccb277de31dc6822ba2b6

mopidy_tidal/library.py
BlackLight: @tehkillerbee this should be fixed in https://github.com/tehkillerbee/mopidy-tidal/pull/60/commits/2adfb83435a9e7196fef85891ca77efb3d84c749

BlackLight:
> **1a**. Artist images are never loaded when opening "Artists" directory (no caching occurs). When clicking directory "My Artist", Images are requested but never displayed. Instead an error is shown.

> **1b**. Playlists images are never loaded when opening "Playlists" directory (no caching occurs). Opening "My Playlists" directory results in partial loading of album art. Only some of the first playlists has an image. The remaining ones are blank.

These should both be fixed by https://github.com/tehkillerbee/mopidy-tidal/pull/60/commits/b92064caa62d4980185ccb277de31dc6822ba2b6

> **2.** Attempting to open "Playlists" (not My Playlists) directory, results in the following (VERY long log message), although playlists get displayed. Error is only shown immediately after mopidy restart. Otherwise, it can be replicated by clicking "refresh" in the Iris GUI. The error is `Expected a list of Track, not [Playlist...`

This should be fixed by https://github.com/tehkillerbee/mopidy-tidal/pull/60/commits/2adfb83435a9e7196fef85891ca77efb3d84c749

> **4** Currently, we have not set the max resolution to use when requesting images. We should consider changing it, as mentioned above. Otherwise we request images/album art that is 1080x1080. It looks like iris does not like images that are too large(?).

This should be fixed by https://github.com/tehkillerbee/mopidy-tidal/pull/60/commits/063fd0ba7cc38a3ccb7e4fb40814ac5cabf1f210

> **5** Browsing "My Albums" result in the following error, followed by the caching itself.

> **6** Browsing My Playlists, result in the following error, followed by the caching itself.

This is actually expected, it's just the error logging trace that is misleading.

It happens because the extension builds URIs such as `tidal:my_artists` or `tidal:my_albums`. While `browse` provides a way to map those results to folders, `lookup` wants a list of tracks to be returned, so it's safe to just skip these entries. I have removed the error log trace from the file.

> **7** While refreshing "My Albums, certain albums cannot be displayed. Instead, an error is thrown:

I have tried to look up the same album (both through ncmpcpp and Iris) and I couldn't reproduce the error. I can't reproduce it when rendering _My Albums_ either. Maybe it's a cache issue? If you have tried several versions of this PR it's likely that the cache folder may be screwed up :) if wiping the cache doesn't work can you try and replicate it when printing `dir(track)` and `type(track)` as well? The `uri` attribute should definitely be there on a `Track` instance. If it's missing on some tracks perhaps it's because is unavailable in the configured market?

fmarzocca:
> This should be fixed

Thank you. I'm going to update to latest commit and I will report any test.

fmarzocca:
I have update to the latest version of better-cache. These are the findings:

This is after a refresh (for every playlist):

`Jun 10 10:50:00 pab mopidy[1056]: ERROR    [Core-17] mopidy.core.library TidalBackend backend returned bad data: Expected a list of Track, not [Playlist(last_modified=0, name='I miei brani di Shazam', tracks=[Track(album=Album(artists=[Artist(name='C. Tangana', uri='tidal:artist:7995155')], name=`

I don't understand this other error:
`Jun 10 10:50:22 pab mopidy[1056]: ERROR    [TidalBackend-6] mopidy_tidal.library <class 'requests.exceptions.HTTPError'> when processing URI 'tidal:track:3579856:70841514:70841515': 404 Client Error: Not Found for url: https://api.tidalhifi.com/v1/albums/70841514/tracks?sessionId=e5947e8d-14f4-45b7-a77e-3c363cac6ccd&countryCode=IT&limit=999`

BlackLight:
> I have updated to the latest version of better-cache. These are the findings:
> 
> This is after a refresh (for every playlist):
> 
> `Jun 10 10:50:00 pab mopidy[1056]: ERROR    [Core-17] mopidy.core.library TidalBackend backend returned bad data: Expected a list of Track, not [Playlist(last_modified=0, name='I miei brani di Shazam', tracks=[Track(album=Album(artists=[Artist(name='C. Tangana', uri='tidal:artist:7995155')], name=`
> 
> I don't understand this other error:
> `Jun 10 10:50:22 pab mopidy[1056]: ERROR    [TidalBackend-6] mopidy_tidal.library <class 'requests.exceptions.HTTPError'> when processing URI 'tidal:track:3579856:70841514:70841515': 404 Client Error: Not Found for url: https://api.tidalhifi.com/v1/albums/70841514/tracks?sessionId=e5947e8d-14f4-45b7-a77e-3c363cac6ccd&countryCode=IT&limit=999`
> 
> 
> 

The second error is either an album that isn't found in a market, or a maximum requests exceeded.

The first one should be fixed. Have you installed the latest version? If it still occurs, you may want to wipe the cache (older cache files may store playlists as list of tracks instead of objects).

fmarzocca:
> If it still occurs, you may want to wipe the cache (older cache files may store playlists as list of tracks instead of objects).

Yes, wiping the cache the error is gone. But wiping the cache is anyway a major PITA because I have to refresh EACH playlist, one by one. Sending a `core.playlists.refresh` doesn't do the job. Do you know a way to automatically refresh all playlists? I thought mopidy-tidal did this on server restart

BlackLight:
> I'm not saying we need this during development but what do you think about a mechanism to automatically clear the cache when the extension version changes?

I don't expect changes in the cached formats to occur that often. While developing the feature and figuring things out the format can obviously change, but now that we've settled for `pickle`-ing native mopidy objects I don't expect many changes that would justify the effort of a version-based cache management logic.

> But wiping the cache is anyway a major PITA because I have to refresh EACH playlist, one by one. Sending a `core.playlists.refresh` doesn't do the job. Do you know a way to automatically refresh all playlists? I thought mopidy-tidal did this on server restart

Manually wiping the cache should no longer be required - we've settled for storing playlists as native mopidy objects instead of lists of tracks, so there shouldn't be any more breaking changes.

When `core.playlists.refresh` is called, the playlist cache is hit if available, and [it will be refreshed](https://github.com/tehkillerbee/mopidy-tidal/pull/60/files#diff-1aca036ed9f15ffb30acef8176fff9ce30d7a238733e0c9de77a9df9d56c4118R32) if the `last_modified` of the upstream object is greater than the cached one (e.g. if the playlist has changes). Otherwise, there should be no reason to force a cache refresh. Note that this works only if you're using my fork of `python-tidal` `0.6.x`, because the standard version doesn't export the `lastUpdated` field.

Depending on the client implementation, however, a mopidy restart may be required in case of updates. Or at least that's the case for ncmpcpp, because the current playlist `lookup` logic doesn't call the upstream API to refresh the last modified date. I could fix this by forcing an API playlist lookup every time a playlist is looked up, but this could impact performance.

fmarzocca:
> Note that this works only if you're using my fork of `python-tidal` `0.6.x`

Got it, thanks. So i have to install your python-tidal 0.6.x. Don't you think it should be set as a dependency on the final mopidy-tidal?

BlackLight:
> Don't you think it should be set as a dependency on the final mopidy-tidal?

I still hope that I won't need to do that.

My PR is a temporary workaround while waiting (for the past two years) for the maintainer of `python-tidal` to finally push `0.7.x`, which is supposed to be a total rewrite of the library.

The maintainer recently promised that 0.7.x should be coming up soon, and that's why I'm holding on that PR: it may soon become irrelevant, and since 0.7.x may also require the rewrite of some parts of `mopidy-tidal`, I don't want to become the de facto maintainer of the stale 0.6.x branch in the meantime.

However, if this PR gets approved before the new `python-tidal` comes out, it's probably worth to add at least a couple of lines to the documentation about the temporary workaround.

fmarzocca:
> I still hope that I won't need to do that.

I fully understand and agree with you. A friend of mine is used to say: "We are on this Earth to suffer". (maybe too much Calvinistic!)

mopidy_tidal/library.py
```diff
@@ -157,104 +151,136 @@ def search(self, query=None, uris=None, exact=False):
...
+                    cache_updates[cache_name] = cache_updates.get(cache_name, {})
+                    if item_type == 'playlist':
+                        # Playlists should be persisted on the cache as objects,
+                        # not as lists of tracks. Therefore, _lookup_playlist
+                        # returns a tuple that we need to unpack
+                        data, cache_data = data
+
+                    cache_updates[cache_name][uri] = cache_data
+
+                tracks += data if hasattr(data, '__iter__') else [data]
```
tehkillerbee, line 251: We should change L251 to the following:
```
if item_type == 'playlist':
    tracks += data.tracks
else:
    tracks += data if hasattr(data, '__iter__') else [data]
```
or make sure tuple is unpacked when no cache miss occurs. Otherwise, we will still get the error:
```
ERROR    [Core-7] mopidy.core.library TidalBackend backend returned bad data: Expected a list of Track, not [Playlist
```

fmarzocca:
Yes. That's what I meant: caching the url, not the image data.


- Sent from my mobile phone. Sorry for typos -

Il Dom 29 Mag 2022, 19:04 Johannes L ***@***.***> ha scritto:

PR #73, opened 2022-09-02
Support tracks in the format tidal:track:<track_id>

### LINKED ISSUE #69: Better track and album URIs
Track URIs are currently expressed in the format `tidal:track:<artist-id>:<album-id>:<track-id>`, while albums are in the format `tidal:album:<artist-id>:<album-id>`.

While this makes it efficient to lookup the parents of an object without performing an additional query, it also makes compatibility with "standard" Tidal URLs a mess.

I had previously set up some automation with Spotify that would allow me to share a track or album from the app/browser, send it via Tasker/Platypush, and it would then play or get added to the queue on one of my RPis. This was possible because mapping from `https://open.spotify.com/track/<track-id>` to `spotify:track:<track-id>` was straightforward, and one can easily leverage the MPD/JSONRPC interface to add stuff.

This isn't that easy with Tidal URIs, given their format. I would first need to perform a query to retrieve the track or the album, and then construct the URI with all the required parts.

We can probably maintain back-compatibility with the current format, but it should be quite straightforward to add a logic that, upon queue/playlist add, just pops the last item of the `tidal:track:` URI and searches for that ID, instead of validating the URI format just to use the last token anyway.

BlackLight:
Closes #69 

A common use-case is the one where you have a script that gets some Tidal URLs and you want to play them or add them to the current track list.

Given the current format of the track URIs (`tidal:track:<artist_id>:<album_id>:<track_id>`), this is only possible by doing a couple of additional API calls on the client side to retrieve the album and the artist associated to the track in order to build the full URI - that's quite an overhead, especially if we don't want to share Tidal API tokens with other scripts.

The "short format" (`tidal:track:<track_id>`) is also supported by Spotify and other backends, and it makes conversions from `https://tidal.com/browse/track/<track_id>` to `tidal:track:<track_id>` much more straightforward.

Note that this PR doesn't change the default format for tracks defined in the backend. It just adds support for the short format on the `library._lookup_track` method (basically, it performs an additional call to retrieve the album and then it does through the normal flow).

### DIFF
--- a/mopidy_tidal/library.py
+++ b/mopidy_tidal/library.py
@@ -472,14 +472,21 @@ class TidalLibraryProvider(backend.LibraryProvider):
         return album.tracks()
 
     def _lookup_track(self, session, parts):
-        album_id = parts[3]
+        if len(parts) == 3:   # Track in format `tidal:track:<track_id>`
+            track_id = parts[2]
+            track = session.track(track_id)
+            album_id = str(track.album.id)
+        else:   # Track in format `tidal:track:<artist_id>:<album_id>:<track_id>`
+            album_id = parts[3]
+            track_id = parts[4]
+
         album_uri = ':'.join(['tidal', 'album', album_id])
 
         tracks = self._album_cache.get(album_uri)
         if tracks is None:
             tracks = self._get_album_tracks(session, album_id)
 
-        track = [t for t in tracks if t.id == int(parts[4])][0]
+        track = [t for t in tracks if t.id == int(track_id)][0]
         artist = full_models_mappers.create_mopidy_artist(track.artist)
         album = full_models_mappers.create_mopidy_album(track.album, artist)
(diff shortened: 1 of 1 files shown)


tehkillerbee:
LGTM!

tehkillerbee (approved):

2e0byo:
This PR doesn't support the 0.6 api (where `session.track` should be `session.get_track`) AFAICS.

Given that, and the fact that setup.py officially requires tidalapi >=0.7, can we just drop all the 0.6 logic?

tehkillerbee:
@2e0byo agreed, let's make it an issue.

BlackLight:
@2e0byo yes, now that the requirements explicitly list `tidalapi >= 0.7.0` I think that it's safe to drop the 0.6.x support logic - that was only supposed to be a crutch anyway before the release of tidalapi 0.7.

PR #76, opened 2022-09-05
Support for playlist edit

### LINKED ISSUE #75: Support for playlist modifications
Implement the `PlaylistProvider` `create`, `delete` and `save` methods.

BlackLight:
Closes #75 

Known limitations:

- Bulk removals may be a bit slow because the current API seems to only support one removal per call
- Additions are always appended to the end of the playlist because that's the only API that seems supported
- Swapping track positions isn't really supported - a first look at the call made by the Tidal client shows a `POST` request to the playlist item URL with the `toIndex` argument, but I haven't found anything like that in the `tidalapi` package


BlackLight:
Any volunteers who would like to test this PR? @tehkillerbee @2e0byo @fmarzocca

I have been doing playlist edits for the past few days without issues, tested both from ncmpcpp, Iris and M.A.L.P. However, all these clients are also compatible with the limitations of the current implementation - only support for "Add to playlist" (i.e. append at the end, not in the middle), and swapping only works with adjacent tracks so we can get away with a DELETE followed by a POST. I'd be interest to know how things work with other clients though.

tehkillerbee:
@BlackLight Definitely. I can only test with Iris, as this is what I currently. If there are other clients you suggest testing with, I'll give it a try.

tehkillerbee:
@BlackLight First initial tests:
1. Adding to existing playlist in Iris does not seem to work; "Action failed (addTracksToPlaylist)
Not supported on this type of object"
2. Creating new playlist and then adding tracks to it works great.
3. Deleting new playlists, existing playlists works, although playlist view must be refreshed manually.
4. Existing playlists are not always shown in the "Add to playlist" dropdown. It seems the playlists must be browsed before they appear.

2e0byo:
@tehkillerbee I *think* 4 is just a limitation of iris in general.  I'm up for testing this, though I too only use iris!  I'll pull it down in a bit.

tehkillerbee:
> @tehkillerbee I _think_ 4 is just a limitation of iris in general. I'm up for testing this, though I too only use iris! I'll pull it down in a bit.

After further testing, I noticed the playlists provided as m3u files show up automatically. So I wonder how they are populated in Iris.
image

BlackLight:
@tehkillerbee I think some work is required on mopidy-iris side as well.

At a first look, it seems that Iris inspect the [`can_edit` attribute](https://github.com/jaedb/Iris/blob/11d92ed262eacdf34ea688036edd27598e7d0d1c/src/js/views/Playlist.js#L64) on the `Playlist` object to tell whether a user can create a playlist of that type, or add/remove tracks into it.

An entry exists for Spotify - one should probably be added for Tidal as well.

Besides that, playlist editing, creation and deletion seems to work quite well over MPD. I mostly use the mopidy-mpd backend to connect to MPD clients, and I have tested playlist edit mostly through ncmpcpp and mpc command-line commands. mopidy-mobile also works by the way, since it's a bit less opinionated than Iris about what playlists can actually be edited.

tehkillerbee:
@BlackLight Since this is mainly an Iris Issue, I think we can merge this PR and add it to the Iris todo list.

tehkillerbee (approved):
Approved, with isort errors

Qbondar (approved):

Issue #78, opened 2022-09-11
Drop all tidalapi 0.6 logic


tehkillerbee:
Cleanup tidalapi 0.6 leftovers. 

_Originally posted by @2e0byo in https://github.com/tehkillerbee/mopidy-tidal/issues/73#issuecomment-1242997999_

PR #82, opened 2022-09-24
Removed legacy code for tidalapi < 0.7 support

BlackLight:
Closes: #78

### DIFF
--- a/mopidy_tidal/backend.py
+++ b/mopidy_tidal/backend.py
@@ -12,13 +12,6 @@ from mopidy_tidal import Extension, context, library, playback, playlists
 
 logger = logging.getLogger(__name__)
 
-try:
-    from tidalapi import __version__
-
-    has_python_tidal_0_7 = True
-except ImportError:  # pragma: no cover
-    has_python_tidal_0_7 = False
-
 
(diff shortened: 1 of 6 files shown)


BlackLight:
@2e0byo it looks like a few tests need to be changed - there are several failures due to the fact that we are now transparently passing `Mock` objects around instead of unpacking their values.

2e0byo:
@BlackLight I don't know how to pr against a pr...

These errors were all just the test suite assuming the old api as default (which worked since the mocked session didn't have the attributes whcih marked it as the new api).

All fixed and passing in https://github.com/2e0byo/mopidy-tidal/tree/remove-tidalapi-0-6-logic

Can you manually pull in that commit (or convert it to a patch and apply or something)?  Otherwise I can pr from my branch and we can merge that: the PR will show up as mine but the commits as yours, since I just forked from you.

2e0byo:
Actually it occurs to me that @tehkillerbee is likely not as silly as I am about only merging green commits.  In which case the simplest would be to merge this PR first, and then mine, repairing the test suite.  I'll open a pr.  But if @BlackLight modifies this PR I'll close mine.

tehkillerbee (approved):
LGTM

Issue #91, opened 2022-10-07
Make cached playlist expiry time on lookup configurable


BlackLight:
The [current playlist lookup/refresh logic](https://github.com/tehkillerbee/mopidy-tidal/blob/master/mopidy_tidal/playlists.py#L242) uses a cached playlist snapshot for 5 minutes.

If a `lookup` is done within that time, then the cached snapshot will be used. Otherwise, the playlist will be refreshed from upstream.

This can be a reasonable trade-off between responsiveness of the UI vs. responsiveness of the application to external changes, but it's not always the best trade-off for everyone. Some users may not modify playlists externally so often, and hence the overhead of the upstream fetch is mostly unneeded. Some instead may want the application to have a high responsiveness to playlist changes, even if that means sacrificing a bit more time on lookup.

We probably shouldn't make assumptions about the trade-off, so this value should be configurable. The default value should probably be zero, i.e. don't refresh a playlist after the first lookup unless the application is restarted (this is a behaviour similar to that of `mopidy-spotify`).

tehkillerbee:
Good idea, this has now been merged. Marking as closed

PR #92, opened 2022-10-07
[#91] Implemented a configurable `playlist_cache_refresh_secs`

BlackLight:
See #91

### DIFF
--- a/mopidy_tidal/__init__.py
+++ b/mopidy_tidal/__init__.py
@@ -30,6 +30,7 @@ class Extension(ext.Extension):
         schema["quality"] = config.String(choices=["LOSSLESS", "HIGH", "LOW"])
         schema["client_id"] = config.String(optional=True)
         schema["client_secret"] = config.String(optional=True)
+        schema["playlist_cache_refresh_secs"] = config.Float(optional=True)
         return schema
 
     def setup(self, registry):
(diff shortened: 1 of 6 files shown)


mopidy_tidal/ext.conf
```diff
@@ -3,3 +3,4 @@ enabled = true
 quality=LOSSLESS
 client_id=
 client_secret=
+playlist_cache_refresh_secs = 0
```
tehkillerbee, line 6: Previously it looks like we used 300 as the default value (hardcoded). Should we keep it here or do we want to use 0 as the default value from now on?

mopidy_tidal/__init__.py
```diff
@@ -30,6 +30,7 @@ def get_config_schema(self):
         schema["quality"] = config.String(choices=["LOSSLESS", "HIGH", "LOW"])
         schema["client_id"] = config.String(optional=True)
         schema["client_secret"] = config.String(optional=True)
+        schema["playlist_cache_refresh_secs"] = config.Float(optional=True)
```
tehkillerbee, line 33: Do we need a Float here or is Int enough?

mopidy_tidal/ext.conf
tehkillerbee: Looks like you have already elaborated on this in #91. Marking this as resolved.

tehkillerbee (approved):

PR #98, opened 2023-03-15
Always add newly created playlists to cache

BlackLight:
If `playlist_cache_refresh_secs` is unset then the playlist won't be automatically added to `_current_tidal_playlists`. We should do it explicitly if we want the backend to be able to immediately get its contents.

### DIFF
--- a/mopidy_tidal/__init__.py
+++ b/mopidy_tidal/__init__.py
@@ -16,7 +16,6 @@ sys.path.append(file_dir)
 
 
 class Extension(ext.Extension):
-
     dist_name = "Mopidy-Tidal"
     ext_name = "tidal"
     version = __version__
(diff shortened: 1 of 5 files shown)


tehkillerbee:
@BlackLight @2e0byo This PR fails all tests; do we need to update the tests as well?

2e0byo:
@tehkillerbee I'll get on this shortly.

2e0byo:
I think there was actually a bug in this PR.  We want to add the tidal playlist, not the mopidy playlist, to `_current_tidal_playlists`. So I think the test suite actually earned its keep.

However once this is fixed we need to fix the test suite, as `create()` now triggers a `refresh()`, so we have to mock `playlist.tracks` (to return nothing, and have a `__name__`).

The attached patch fixes this.
[0001-fix-cache-tidal-playlist-not-mopidy-playlist.txt](https://github.com/tehkillerbee/mopidy-tidal/files/11192220/0001-fix-cache-tidal-playlist-not-mopidy-playlist.txt)

BlackLight:
Thanks for the fix @2e0byo!

tehkillerbee (approved):

PR #99, opened 2023-03-15
Temporary fix for the missing user picture error on tidalapi

BlackLight:
tidalapi recently started breaking after logging in because of the missing `picture` field on the user JSON response (see https://github.com/tamland/python-tidal/issues/129).

I have [pushed a PR](https://github.com/tamland/python-tidal/pull/130) to quickly fix the issue, but it may take a while to get merged (the project has a quite long queue of pending PRs).

In the meantime, as a workaround, this PR makes the project depend on my fork (which includes the fix) instead of the latest published release.

### DIFF
--- a/mopidy_tidal/__init__.py
+++ b/mopidy_tidal/__init__.py
@@ -16,7 +16,6 @@ sys.path.append(file_dir)
 
 
 class Extension(ext.Extension):
-
     dist_name = "Mopidy-Tidal"
     ext_name = "tidal"
     version = __version__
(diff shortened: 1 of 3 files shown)


tehkillerbee:
@BlackLight I would prefer if we do not change to a fork of python-tidal, since it would also require updating the mopidy-tidal pypi release, and then we also have to change it back again when the PR is merged. But I am worried that it will take a while for python-tidal to be updated.

For now, I would suggest that we install from your fork - manually. If the PR is not merged soon, then we will switch to your fork. What do you think?

tehkillerbee (approved):
LGTM

tehkillerbee:
It looks like the PR has not been merged yet. I will switch to your fork for now, then lets switch back when the PR has been processed.

Issue #111, opened 2023-05-01
Add integration tests.


2e0byo:
Apparently
> Deprecated field ‘track’ with the goal of removing it in the next major release, use ‘track_name’ instead. Backends should support both track and track_name until they require a version of Mopidy where track has been removed. [src](https://docs.mopidy.com/en/latest/changelog/#v3-3-0-2022-04-29)

I don't know if we do this, but I do know that we don't have a test for it.  We need integration tests, preferably against a few versions of mopidy, with `-Werror` so we fail on deprecations.

@tehkillerbee  this would be slightly easier to write (at least for me) if we were using poetry.  If you'd be up for that I'll do it in the PR.  But I'm perfectly happy to figure out how to do it without.

2e0byo:
Ta! I promise not to do anything stupid.  Well, not intentionally anyhow ;)

2e0byo:
Curiously having re-read mopidy code I don't think there *is* any way to trigger this deprecation warning, since we've overriden the method.  I'm just going to fix it for now, but integration tests are on the way anyhow.

PR #113, opened 2023-05-07
Ci/integration tests

2e0byo:
This PR migrates to poetry and adds integration tests.  It actually doesn't test for deprecation warnings (yet) as I wasn't able to trigger any.  The warning I noticed can't be caught I think as we override the method, so I've just fixed it as part of the PR.

These tests are 'user stories' in the web world. Currently there are two:

- user loads up default config.  Mopidy pauses and generates a tidal login url.
- user loads up lazy config.  Mopidy does not generate a tidal login url.  User accesses tidal (via `mpc list artist`).  Mopidy generates a tidal url.

FWIW I think you'll find the user-story style tests with pexpect easier to read.  I've never used it before and was pleased how easy it was.

We could have other user-story tests if we had a mock tidalapi, but tbh I think we've got enough users.  This just stops people (=me) PRing hopelessly broken builds.

I've deliberately built this at the highest level, i.e. actual subshells.  It would be possible to call into mopidy ourselves, perhaps by instantising `core`, and get better access to the internals.  But I don't want to learn that much about Mopidy tonight.

We test against all major-numbered supported Mopidy versions + latest upstream git and all major-number supported python versions.  With the git test we should get some kind of feedback against upstream.  Thus I've set this to run every week.

# Poetry

I feel I owe an explanation of poetry as I'm suggesting my preferred tooling.  I've updated the readme, but to setup the venv you just do `poetry shell`.  If you want to be able to run mopidy from inside the venv, you can either use system-site-packages like before or just install pygobject (which I prefer, as venvs are supposed to decouple stuff, and we break that as soon as we use system-site-packages). Unfortunately poetry's support for system-site-packages is currently broken... there's a hack in the readme/makefile but <sup><sub>it involves making the venv the old way</sub></sup>.  Ahem.

I *still* think we should move to poetry:

- one config file!
- all the cool kids use it!
- easy releases to pypi!  (just run `poetry release` and it will do the rest)
- really easy dependency management
- reproducible builds + dev environment (the devtools are installed inside the venv, so we have the same versions)

Still I don't want to break anyone's workflow.

# PS

I few other things have sneaked through.  I'm currently renaming tests with the sentence-style names, can't remember if I've mentioned that  before.  Personally I think it adds a lot of clarity.  I mean to base my next PRs off this whenever they land, so if you are happy with everything just merge the most recent.  But we can keep discussion of this here.


2e0byo:
Closes #111.  Also slightly improved the ci and added caching, which saves a whole *two seconds* on average on the test run.

.github/workflows/test.yml
```diff
@@ -11,31 +11,31 @@ jobs:
...
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+          cache: 'poetry'
+      - name: Install mopidy-tidal
+        run: |
+          poetry install
       - name: Lint
         run: |
-          pip install isort black
```
tehkillerbee, line 26: So if I understand correctly, poetry completely takes care of these dependencies now?

.gitignore
```diff
@@ -1,13 +1,176 @@
-*.egg-info
-*.pyc
-*.swp
-.cache/
-.coverage
-.tox/
-MANIFEST
+# Created by https://www.toptal.com/developers/gitignore/api/python
```
tehkillerbee, line 1: Neat! That will come in handy at work!

README.md
tehkillerbee, line 1: We might consider splitting up our README now that the development section has become more detailed.

Perhaps it is time to split it into multiple markdown files. I've wanted to do that with the CHANGELOG as well.

For now, lets merge it into the README, I will create a new issue for a README cleanup.

tehkillerbee (approved):
Looks great! Only minor comments that can be handled in a future issue.

.github/workflows/test.yml
2e0byo: Yes!  They're in the dev group installed by default but not in the build

README.md
2e0byo: Or we could have a really small static site?  something like [this](https://2e0byo.github.io/gallica-autobib/) (shameless plug for one of my own projects).  Or even readthedocs, although I'm not a *massive* fan of sphinx in small projects: I'd rather just read the source.

2e0byo: oops, moving to issue.

PR #116, opened 2023-05-09
Feat: Login track hack

### LINKED ISSUE #89: Display login auth code / login status somewhere else than console.
Hi,

When Mopidy-Tidal starts up, it shows the IP address and network port on a small st7789 screen with mopidy-pidi. before it starts, the screen is blank while waiting for the link.tidal.com/xxxx screen. Is it possible to show the output of "journalctl -u mopidy | tail -10" to that screen during the login prompt timeframe?

2e0byo:
This PR implements a new login method, 'HACK' (which is not the default).  In HACK mode Mopidy-Tidal intercepts all user actions, and if not logged in, returns a fake object (artist, album, track, ref, playlist, etc) with instructions on how to log in.

All image queries return a QR code with the correct link.

Additionally, if the user *plays* a track, a text-to-speech service is used to read out the instructions, because if you're going to have a hack you might as well have an *impressive* hack.

I didn't want to spaghettify our logic for 0.01% of our runtime, so I've implemented everything as two decorators inside `login_hack.py`.  Unfortunately we need to block all kind of functions with different signatures and different return signatures.  Consequently the introspection logic is... exciting.  If you're not used to seeing immediately executed functions and recursion in python you might be forgiven for thinking it was JS or TS.  But it works ;)  Also, if you haven't just emerged from deep in the bowels of python's type system you might find things like `get_args` a tad odd.  Type objects are bare classes constructed with [] rather than (), but they're actually legitimate objects and we can introspect them, modify them, and even construct them dynamically.

I think the implementation could probably be made a little less exciting---I solved the general problem rather than the specific in most cases, partly for fun, and partly as I didn't know what I'd end up needing.  I've since trimmed some branches I couldn't get tests to cover i.e. made some assumptions about real-world cases.  I *think* these are all correct (and we have 100% line and branch coverage, and I've tested with iris and mpc), but if anything shows up I'll fix it.

If anything does break, setting `login_method = BLOCK` will cause the decorators to pass through.

In one instance there is no way to infer the correct type and we have to use a doubled decorator (i.e. `@login_hack(type)`) to pass it through.  But it's still only one line.

# Breaks

Mopidy < 3.3  Deps bumped accordingly.  I don't think we meant to support old mopidy anyhow.
Python < 3.9  (When we use 3.10 I can rewrite with pattern matching... which would have been nice.)

# Security

Both text-to-speech and qr use public apis.  We could ship the latter easily (the former is a waste of space for a gimmic).  Since the url is only any good with a valid tidal login I don't think this matters.

# Caveats

Anything the user does to get the magic objects will need to be redone afterwards.  Mopidy just doesn't have the concept AFAICS of telling a frontend to refresh itself.

Fixes #89.


2e0byo:
I was going to add docs, but I'll wait till the other issue is resolved.

I've done some refactoring along the way: pathlib throughout, fixed the `#noqa` and so on.  Mypy isn't passing yet though.

mopidy_tidal/backend.py
```diff
@@ -71,13 +92,12 @@ def on_start(self):
...
         self._active_session = Session(config)
-        if not self._config["tidal"]["lazy"]:
+        if not self.lazy_connect:
             self._login()
 
     def _login(self):
         # Always store tidal-oauth cache in mopidy core config data_dir
-        data_dir = Extension.get_data_dir(self._config)
-        oauth_file = os.path.join(data_dir, "tidal-oauth.json")
+        oauth_file = self.data_dir / "tidal-oauth.json"
```
tehkillerbee, line 100: Is this syntax preferred over os.path.join?

mopidy_tidal/login_hack.py
```diff
@@ -0,0 +1,304 @@
...
+    base_type = get_origin(target_type)
+    if base_type is dict:
+        return base_type[get_args(target_type)[0], inner_type]
+    if base_type is list:
+        return base_type[inner_type]
+    else:
+        return inner_type
+
+
+voice_rss_api_key = "eb909fe9f2ce403bb7209de172d096f1"
```
tehkillerbee, line 249: Do we want the API key directly in the sources or should we add it as a secret?

mopidy_tidal/backend.py
2e0byo: well it is by me... the difference is that `self.data_dir` is now a `Path()` object, so we can use pathlib stuff on it. One of those is the overriden `__div__` so you can write paths out with `/`.  I cherry-picked [this refactor](https://github.com/tehkillerbee/mopidy-tidal/pull/116/commits/e30aabd74920bb9a2bade9710ff39ab876d7d8a7) from a stalled PR improving the cache, I think because I needed to make/delete dirs and it was easier with pathlib.

But I personally think the pathlib syntax of `path_obj / path_obj / path_obj / "str2"` is neat, and easier to read than `os.path.join`.  (There's no decent os.path replacement I know of for `Path().mkdir(exist_ok=True, parents=True)`, so we were actually casting strs to `Path`s every time we needed to make a directory anyhow.)  The other pathlib way of doing it is `Path(path_obj, path_obj, path_obj, "str", "str")`, but that's too magical in my book---a newcomer isn't going to know that the constructor basically calls `os.path.join` on `*args`, so I try to avoid using it, even preferring string interpolation.  Whereas `/` is so natural you don't even notice it's doing something clever until you think about it.

Thoughts?

mopidy_tidal/login_hack.py
2e0byo: Thing is it has to be used by the *instance*, so we have to ship it.  I deliberately picked a service which didn't require a credit card and has a free tier, so I'm not that worried about exposing it.  If you like I could obfuscate it a bit (e.g. use base64) to defend against automated scraping.  I couldn't think of a good solution though and did try to find a service which didn't require an api key.

2e0byo:
I've been madly busy with my thesis (due on wednesday) but I mean to get back after that and see if I can reproduce the problem with the 2nd time not working reported in the linked issue.

mopidy_tidal/backend.py
tehkillerbee: > But I personally think the pathlib syntax of path_obj / path_obj / path_obj / "str2" is neat, and easier to read than os.path.join

Agreed, its easier to read the path when written in that way. Its probably also a better idea to use a Path() object instead.

tehkillerbee:
> I've been madly busy with my thesis (due on wednesday) but I mean to get back after that and see if I can reproduce the problem with the 2nd time not working reported in the linked issue.

No worries at all. Good luck!

tehkillerbee:
Finally, I had time to do some testing. It seems authentication is not working properly when visiting the link displayed as the track name. Instead, after visiting the link as instructed through mopidy web, refreshing results in a new link is displayed again. 

I have tried refreshing right after visiting the link and also restarting mopidy without refreshing mopidy web. Both results in another login URL displayed. Have you seen this behaviour?

I have set the login lazy to TRUE and login_method to HACK

I do not get anything in the logs either, so it is difficult to know what goes wrong. Perhaps, we should keep the login info and related prompts in the logs?

2e0byo:
I've not had that; I'll see if I can reproduce this evening.

2e0byo:
Okay, I found and fixed the bug and it should actually log in! oops.  Tested this end.  I'm not sure how I missed that.  We were never saving the credentials anyhow.

However this is very clearly a problem with the testing so I'll add tests with a mocked return.

I've also fixed the playlist return type so mock playlists should work again.

tehkillerbee:
@2e0byo Great to hear, I will test tonight and hopefully merge it as well. It's a very cool feature, nonetheless. :)

I see there is a merge conflict in poetry.lock. Do you know what causes this and how to resolve it?

2e0byo:
I'll fix it. Just need to prefer master's.

tehkillerbee (approved):
Sorry for the delay. Everything appears to work now. Great work!

PR #131, opened 2023-08-05
Feature: Add HI_RES support, handle missing images gracefully

tehkillerbee:
This PR adds the following features:

- Added HI_RES support
- Updated readme with HI_RES details
- Add log message if HI_RES quality is not available for selected track
- Gracefully handle URIs with no images associated to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,15 @@ Mopidy Extension for Tidal music service integration.
 ## Installation
 First install and configure Mopidy as per the instructions listed [here](https://docs.mopidy.com/en/latest/installation/). It is encouraged to install Mopidy as a systemd service, as per the instructions listed [here](https://docs.mopidy.com/en/latest/running/service/). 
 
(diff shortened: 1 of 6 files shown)


2e0byo:
hmm I think the tests might need looking at... ;)

2e0byo:
@tehkillerbee I'll have a look at fixing this up in the next few weeks if you haven't by then.  We should get the tests passing: the integration failure is probably indicative of a real regression somewhere in the public api surface. The unit tests might just need to match the api better: the problem with using mocks everywhere.

tehkillerbee:
> @tehkillerbee I'll have a look at fixing this up in the next few weeks if you haven't by then. We should get the tests passing: the integration failure is probably indicative of a real regression somewhere in the public api surface. The unit tests might just need to match the api better: the problem with using mocks everywhere.

I will try to fix the tests but it might take a while. If I cannot get it working, I might end up merging this PR so we can get HI_RES support added. Then I will fix the tests in a separate PR. Not ideal, of course.

Issue #137, opened 2023-09-03
Incorrect tests for nonexistent objects


2e0byo:
See e.g. https://github.com/tehkillerbee/mopidy-tidal/blob/33b41e1f15a425fd983c172bacd48147a30f1fe2/mopidy_tidal/library.py#L395

This check is incorrect as tidalapi never returns anything falsy, but *does* raise an exception if an invalid album is accessed.

2e0byo:
The recent typing work will actually catch these bugs, so that's good

tehkillerbee:
Fixed in https://github.com/tehkillerbee/mopidy-tidal/pull/151

PR #138, opened 2023-09-03
test/rewrite test suite

2e0byo:
@tehkillerbee here's a start on the work of refactoring the test suite to be (i) more correct and (ii) easier to maintain.

So far I've not fixed the broken tests, but I've focused on making the current tests descriptive.  This is a good exercise and led to finding a few bugs, like #137, where before I'd blindly asserted our behaviour without checking it made sense.

As well as general criticism I'd really appreciate feedback on how clear the new tests are, and how obviously they set out what they're testing.  The new tests are in `test_library.py`: the rest is just cleanup.

We're always going to have some 'magic', but I've taken a few steps to minimise surprises:

- prefer factory functions to predefined test data (so `make_tidal_track` over `tidal_tracks`) -> the test is more self-contained, and it's obvious where the data is coming from
- move funny logic into conftest.py
- set specs on all the mocks (basically making them look like the class they're mocking)

I've also fixed a few bugs and done some refactoring.

I mean to chip away at this, but when you're happy with things feel free to merge this pr in (we might as well improve incrementally).

I mean to:

- continue the process for all tests
- fix broken tests and get our coverage up to 100% again
- refactor some of the code which is rather verbose (mostly along the lines of the ifs I refactored in this)
- split out some of the tests of e.g. caching, which have too many assertions
- dig up some PRs I have lying around as branches to start adding in track caching (which I've nearly finished and am using locally)

over the next few months in odd moments, so hopefully there'll be a steady drip of small PRs rather than one massive chunk.


2e0byo:
sorry, the *new* tests are kindof buried in trivial changes.  You probably want to look over here:
https://github.com/tehkillerbee/mopidy-tidal/blob/test/rewrite-test-suite/tests/test_library.py#L147-L297

This is the style I'm trying to roll out everywhere.

2e0byo:
@tehkillerbee  this rewrite is 64% complete by files and probably 75% complete by lines.  I need to fix up 3.9 (if we still support it) but it might be worth having a look and maybe merging it in in this state.

Two things are still needed to hit the original goal of making the tests obvious:

1. I need to finish the other files, i.e.:
  - [ ] test_image_getter.py
  - [ ] test_library.py # partially done
  - [ ] test_playback.py
  - [ ] test_playlist_cache.py
  - [ ] test_playlist.py
  - [ ] test_search.py
2. mocking python-tidal needs to get easier.

I'm inclined to attack 2. first. It would be cool if we could just do `from tidalapi.testing import MockTrack; MockTrack("name")` and it would Just Work (TM).

Off the top of my head a slight refactor over at python-tidal (dropping `None` + hydration) would make this easier.  In any case I'm inclined to work on that first and then come back and finish this here.  It doesn't make sense that `mopidy-tidal` is growing code to make mock tidal objects, which everyone who uses tidal can benefit from.  It also doesn't make sense that we have to use mocks (because we don't want network access) when most of the object is just data, but I'll raise that over there.

I don't know what you want to do about this PR, but it's a lot of red/green to review. It might be worth just looking at the new test files and highlighting what doesn't make sense, rather than looking at the tests side-by-side.  I'm happy to clarify or re-write anything which isn't clear.

2e0byo:
The test failures are mostly spurious:
- all the integration tests against upstream mopidy have failed, because it now requires python >= 3.11 (poetry takes a strict approach to python versions which is highly annoying, so it refuses to install even on 3.11 because we claim to support old python... oh well)
- codecov/project complains that I've xfailed rather than fixing broken tests, slightly reducing our coverage
- codecov/patch complains that some of the PR isn't tested, which is because... it's testing.

tehkillerbee:
Hey, good to see you've had time to work on this.

> mocking python-tidal needs to get easier.
Agreed, sounds like this is necessary before proceeding with this.

In any case, I'll look through the changes and review them.

It would be nice to get it merged, especially before next mopidy-tidal release. There hasn't been one for a while, but I am planning to release it when I have gotten Hires/HLS stream playback working.

2e0byo:
Thanks for your time!  Sorry this PR's been open for so long...

mopidy_tidal/library.py
```diff
@@ -222,50 +222,55 @@ def browse(self, uri) -> list[Ref]:
...
-            return ref_models_mappers.create_tracks(
-                self._get_mix_tracks(session, parts[2])
-            )
-
-        if nr_of_parts == 3 and parts[1] == "page":
-            return ref_models_mappers.create_mixed_directory(session.page.get(parts[2]))
-
-        logger.debug("Unknown uri for browse request: %s", uri)
-        return []
+        try:
```
tehkillerbee, line 225: Neat, I like this way much better!

mopidy_tidal/library.py
```diff
@@ -222,50 +222,55 @@ def browse(self, uri) -> list[Ref]:
...
+            elif type == "mix":
+                return ref_models_mappers.create_tracks(
+                    self._get_mix_tracks(session, id)
+                )
+
+            elif type == "page":
+                return ref_models_mappers.create_mixed_directory(session.page.get(id))
+            else:
+                return []
+        except ValueError:
```
tehkillerbee, line 268: Double check this exception type is still used after my latest additions to tidalapi

pyproject.toml
```diff
@@ -50,10 +50,9 @@ tidal = "mopidy_tidal:Extension"
 
 [tool.pytest.ini_options]
 markers = [
-    "gt_3_7: Mark a test as requiring python > 3.7.",
-    "gt_3_8: Mark a test as requiring python > 3.8.",
-    "gt_3_9: Mark a test as requiring python > 3.9.",
```
tehkillerbee, line 55: Are we not supporting 3.9?

tests/test_cache_search_key.py
```diff
@@ -3,41 +3,22 @@
...
+    assert SearchKey(**params) == SearchKey(**params)
 
-    d1_sk = SearchKey(**d1)
-    d2_sk = SearchKey(**d2)
+    assert hash(SearchKey(**params)) == hash(SearchKey(**params))
 
-    assert hash(d1_sk) == hash(d2_sk)
 
+def test_hashes_of_different_objects_are_different():
+    key_1 = SearchKey(exact=True, query=dict(artist="Arty", album="Alby"))
```
tehkillerbee, line 14: :+1:

tests/test_extension.py
```diff
@@ -15,24 +15,30 @@ def test_get_default_config():
...
     ext = Extension()
 
     schema = ext.get_config_schema()
 
-    # Test the content of your config schema
-    assert "quality" in schema
-    assert "client_id" in schema
-    assert "client_secret" in schema
-    assert "lazy" in schema
+    assert set(schema.keys()) == {
```
tehkillerbee, line 24: The original might be easier to understand on first glance. But I guess your changes ensure that a forgotten key will be caught.

tests/test_library.py
```diff
@@ -1,303 +1,360 @@
...
-    session = backend.session
-    session.user.favorites.albums = tidal_albums
-    mocker.patch("mopidy_tidal.library.get_items", lambda x: x)
-    assert tlp.browse("tidal:my_albums") == [
-        Ref(name="Album-0", type="album", uri="tidal:album:0"),
-        Ref(name="Album-1", type="album", uri="tidal:album:1"),
-    ]
 
+class TestSearch:
+    def test_defers_to_tidal_search(self, library_provider, mocker):
```
tehkillerbee, line 50: Defer in this context = ?

tehkillerbee (approved):
I think it looks pretty good. I like the structure of the new tests added to test_library.py. It looks relatively easy to understand what is going on.

Only minor comments. I think we can get this merged in and then deal with the comments in a later MR.

2e0byo:
Thanks!  I'll go through after work and cleanup.

pyproject.toml
2e0byo: It's more that the marker isn't useful as 3.9 is the lowest python we support (so this would just mark everything).

Incidentally mopidy's develop branch now requires python 3.11, although the last release still claims to support 3.7.

tests/test_extension.py
2e0byo: yes, it was more about ensuring that we *only* have the keys we think we have.  This test is largely pointless: it's mostly cheap insurance against forgetting to add a config in one place but requiring it in another.

tests/test_library.py
2e0byo: I use `defer` when I test that some functionality is just handed over to another unit we test somewhere else.  So if `foo(x)` just calls `bar(x)` (and the fact that `foo` calls `bar` is part of the contract) the only behaviour we can assert about `foo` is that it hands the right data over to `bar` and then returns the result.

Concretely here, search is simply a wrapper around `python_tidal`'s search.  The only thing it should do is format the query, hand it to `python_tidal`, format the result and then hand it out. The actual searching is deferred to the underlying library, and we shouldn't try to test it.

It's just a word I use; don't think you'll find any blog posts on it. There's probably a proper term for it.

Contrast this to testing that e.g. `TestBrowse:test_invalid_uri_returns_empty_list`.  That behaviour is part of the spec of `browse()`, so we just hand in an invalid uri and check that we get an empty list back---the particular details of how that happens aren't part of the spec and shouldn't be tested.  If `browse` call some internal function, that's its business.

tests/test_library.py
```diff
@@ -1,303 +1,360 @@
...
+        assert res == {"Title-0 [TIDAL]", "Title-1 [TIDAL]"}
+
+    def test_returns_empty_set_when_no_match(self, library_provider):
+        assert library_provider.get_distinct("nonsuch") == set()
+
+    def test_returns_empty_set_when_no_match_with_query(self, library_provider):
+        assert library_provider.get_distinct("nonsuch", query={"any": "any"}) == set()
+
+    @pytest.mark.parametrize("field", ("artist", "track"))
+    def test_query_ignored_when_field_is_not_album_or_albumartist(  # TODO why do we do this?
```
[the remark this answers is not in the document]
2e0byo: I'd rather mark the behaviour for deletion.  But there's some careful logic on it so we should figure out why it was written first.

tests/test_lru_cache.py
```diff
@@ -8,176 +8,233 @@
...
+        cache = LruCache(max_size=8, persist=True, directory="cache")
+        cache.update({"tidal:uri:val": "hi", "tidal:uri:otherval": 17})
+        cache.cache_file("tidal:uri:val").unlink()
+        del cache
+
+        new_cache = LruCache(max_size=8, persist=True, directory="cache")
+        new_cache.prune("tidal:uri:otherval")
+        new_cache.prune("tidal:uri:val")
+
+    def test_migrates_old_filename_if_present(self, lru_disk_cache):
```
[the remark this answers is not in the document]
2e0byo: agreed, probably in the next major release.

The worst that will happen is a cache miss (+ stale files on disk).

2e0byo:
@tehkillerbee I'd like to check that error hasn't changed upstream, then this is good to merge from my POV.

tests/test_library.py
tehkillerbee: Thanks for the detailed explanation, that makes good sense.

tests/test_lru_cache.py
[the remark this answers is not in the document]
tehkillerbee: https://github.com/tehkillerbee/mopidy-tidal/issues/150

tehkillerbee:
@2e0byo If you have time, you can rebase to latest master then I will merge this PR before v0.3.5 release :+1:

2e0byo:
btw `[str]` (in backend) should be `list[str]`. Annoyingly `[]` doesn't mean 'list literal' in a typing context.  Fwiw I've made it a tuple anyway, as it might as well be immutable.

2e0byo:
Lots of tests need fixing up with the latest changes; I'll try to get to it soon.  This is rebased, however.

2e0byo:
login=hack is broken, so if you want to release pronto you should remove it.  I see there's now an AUTO method: if you get to this soon, what does that do? (else I can just read the code).  We may not need the hack login if we have something better.

tehkillerbee:
@2e0byo are you sure? Login hack worked fine last time I tested, after refactoring. I tested with PKCE but perhaps it's broken with Oauth.. :(

Auto=hack, I just added auto as an alias as it sounds more user friendly.

tehkillerbee:
Thanks for the rebase. No rush with the tests, I'll make a branch and I'll try to take a look at the broken tests.

2e0byo:
BTW if you do get round to looking at tests, I think the integration tests are the first thing to fix.  They're all broken because the log output has changed, but should be very easy to fix.  That would be a nice way to confirm that everything does in fact work.

PR #142, opened 2023-09-27
Use full_name instead of name to better identify tracks in mopidy

vaosilva:
Won't expand much further than the title, using the `tidal_track.name` as the `mopidy_track.name` results on a very confusing experience, especially on Single EPs with an original track and multiple remixes of it, being presented all with the same name.

I checked the python-tidal repo and now we can get a `full_name` containing the track `version` field on it: https://github.com/tamland/python-tidal/blob/1d2ae41967368a2ce2d0880b0d28adf5b2dbe553/tidalapi/media.py#L177

Took advantage of that and updated here a simple spot and the tests affected.

### DIFF
--- a/mopidy_tidal/full_models_mappers.py
+++ b/mopidy_tidal/full_models_mappers.py
@@ -63,7 +63,7 @@ def create_mopidy_track(artist, album, tidal_track):
     track_len = tidal_track.duration * 1000
     return Track(
         uri=uri,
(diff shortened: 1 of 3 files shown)


tehkillerbee (approved):

tehkillerbee:
Thanks for the PR, much appreciated. Code LGTM, Merged!

Issue #147, opened 2024-02-07
Fix HI_RES support


tehkillerbee:
While HI_RES support was added last year, actual playback of HiRes albums was not possible since PKCE login was necessary before getting access to HiRes streams. This feature has now (almost) been added to tidalapi so we can finally get proper HiRes support in mopidy-tidal

This task is still in progress (see https://github.com/tehkillerbee/mopidy-tidal/tree/feature/hires-playback) but HiRes playback does work in this branch. Proof:
```
Feb 07 21:57:17 mopidy[11915]: INFO     [TidalBackend-6 (_actor_loop)] mopidy_tidal.library Returning 12 tracks
Feb 07 21:57:17 mopidy[11915]: INFO     [TidalBackend-6 (_actor_loop)] mopidy_tidal.playback TIDAL uri: tidal:track:24905:77646169:77646170
Feb 07 21:57:17 mopidy[11915]: INFO     [TidalBackend-6 (_actor_loop)] mopidy_tidal.playback MimeType:application/dash+xml
Feb 07 21:57:17 mopidy[11915]: INFO     [TidalBackend-6 (_actor_loop)] mopidy_tidal.playback Starting playback of track:77646170, (quality:HI_RES_LOSSLESS, codec:flac, 24bit/192000Hz)
```

- [x] Modify playback to use (HLS) stream instead of direct URLs.
- [x] Handle PKCE login nicely (Currently session .json must be generated manually and copied into the correct folder)

tehkillerbee:
Fixed in #151

PR #149, opened 2024-02-13
New Model Mapping

quodrum-glas:
This is not ready to merge, as it requires more typing and testing must be integrated.
But it is worth implementing as is faster and easier to maintain
I don't have the time to do more, I stripped down testing and typing and use it from my [fork](https://github.com/quodrum-glas/mopidy-tidal/tree/quodrumglas034) (or [this](https://github.com/quodrum-glas/mopidy-tidal/tree/quodrumglas) without merging many commits).
I would not part from it. I tested it extensively from phone app.

This is stripped down from my fork where I have more changes, in order to keep it just to the new model mapping.
The lack of time to invest to make this a true pull request, means I keep my changes obscure in my own space.

If you think is worth it, maybe there are other contributors willing to do the necessary testing code, typing, login_hack decoration, etc
If this gets integrated, then I will contribute more to the master

PS There might be some bugs, but that would come from the tidalapi I have a update many commits behind [here](https://github.com/tamland/python-tidal/compare/master...quodrum-glas:python-tidal:quodrumglas)
At least the [one](https://github.com/tamland/python-tidal/compare/master...quodrum-glas:python-tidal:quodrumglas#diff-d1dd66bd1f77ea832914d423a74708b0eb185213d8eb4c12bc7568c8214d2986) on page recursion is still happening


tehkillerbee:
@quodrum-glas Thanks, your contribution is much appreciated, even if it is not a complete PR.

Of course, this will require quite a bit of work to implement your changes but it looks like it could be worth it in the long run. Perhaps @2e0byo or @BlackLight could be interested in having a look. Otherwise, I'd take a look when the latest HiRes features are added to this release of mopidy-tidal

> This is stripped down from my fork where I have more changes, in order to keep it just to the new model mapping.

Feel free to link to this other fork, if you have more relevant additions that you think are relevant to add to mopidy-tidal

> PS There might be some bugs, but that would come from the tidalapi I have a update many commits behind 
Perhaps it might be a good idea to add a similar PR to python-tidal so your bugfixes can be added there so I can get them merged in.

2e0byo:
@quodrum-glas thanks for this, it's really good to see.

Couple of high level comments:

- it's really nice to see model conversion in classmethods not functions.  I actually chose to stop rewriting the tests when I got to that part recently in the hope that this would happen first.

mopidy_tidal/library.py
```diff
@@ -167,272 +49,79 @@ def get_distinct(self, field, query=None) -> set[str]:
                         apply_watermark(a.name)
                         for a in self._get_artist_albums(session, artist_id)
                     ]
-            elif field in {"track", "track_name"}:
+            elif field == "track":
```
BlackLight, line 52: Duplicate condition check?

mopidy_tidal/library.py
```diff
@@ -1,155 +1,37 @@
...
+    root_directory = Ref.directory(uri=str(URI(URIType.DIRECTORY)), name="Tidal")
 
-    @login_hack(passthrough=True)
-    def get_distinct(self, field, query=None) -> set[str]:
+    def get_distinct(self, field, query=None):
         from mopidy_tidal.search import tidal_search
 
-        logger.debug("Browsing distinct %s with query %r", field, query)
-        session = self._session
+        logger.info(f"Browsing distinct {field!s} with query {query!r}")
```
BlackLight, line 22: Nitpicking, but I think (at least according to Black's style conventions) the `log.info("msg: %s", var)` notation is preferred over `log.info(f"msg: {msg}")` - at least for logging functions.

2e0byo: (the reason for this btw is that f-strings are eagerly evaluated, but a logging function might never produce any output. Using `%` pushes the formatting down the stack and (in theory) avoids doing it if the log level suppresses the output)

tehkillerbee:
>  Maybe is something to help your 2nd part of [Fix HI_RES support #147](https://github.com/tehkillerbee/mopidy-tidal/issues/147) ?

@quodrum-glas Thanks for the suggestion. I ended up implemented it as you suggested, using a local HTTP server based on your code. It works rather well! I had to rewrite most of the relevant backend code but it works and it can also be used for OAuth login.

mopidy_tidal/library.py
quodrum-glas: That is correct. Don't consider this change. I was trying to extract only the model mapping change and other things were included. And I was wrong to change them, anyway.

quodrum-glas:
@2e0byo 
Thank you for taking a look at this. Sorry for late reply, I did not expect the attention. Also was on some time off.

The pull request is only to raise awareness on this model mapping that is quite responsive.
It is true, I have not used the master version in a long time to compare any advancements that happened there.

Also this model mapping adds more items. It adds pages as per native app browsing, etc.
To actually implement it, requires some effort to merge the filtered core from this request, improve, and then add tests, lint it, etc. I can contribute to this within a limited time%.

The cache is defined centrally in one place and can be adjusted easily to best suited balance.
The CacheTools has also TTLCache (as implemented at playlist module).
I considered that, for example, cache by uri (so track info from id) there won't be any changes server side.

There were a lot of re-requests for data. Especially for track data. Quite expensive.

I welcome the point about `pydantic`.

I believe the value here is that I tried to match the need for server data with browsing behaviour and frontend structure (as in only request what is needed at the time and no further content without explicit request)
, as well as avoiding re-requesting data from server if was supplied earlier in one form or another. (due to incompatibilities in matching exactly server data structure to frontend Mopidy).
And that's effectively what I try to "sell" here :) ... the core idea in models.py module and the implementation in playlists.py and library.py

This has lived with me for a while and naturally can be improved already.

mopidy_tidal/library.py
quodrum-glas: This might be a consequence of trying to merge something I did not change but remained as part of much earlier fork.
Effectively is a "roll back" :)

quodrum-glas:
To test this implementation and see the actual outcome, one can go to this [ReadMe](https://github.com/quodrum-glas/mopidy-tidal/tree/quodrumglas-dev?tab=readme-ov-file#getting-started) and use the command as stated there:
```
sudo python3 -m pip install -U 'git+https://github.com/quodrum-glas/mopidy-tidal.git@quodrumglas-dev#egg=Mopidy-Tidal'
```
As you know...
```
python3 -m pip uninstall mopidy-tidal
python3 -m pip uninstall tidalapi
```
...must be done beforehand

Or use `--force-reinstall` but will reinstall all dependencies.

`tidalapi` must be reinstalled from my branch to address a couple of small bugs, especially at Page class
A pull request for that fix I submitted here: https://github.com/tamland/python-tidal/pull/236

Another login is required (web server) but won't overwrite the stored session from master

2e0byo:
@quodrum-glas minor point on the uri: it *shouldn't* change, but tidal do actually change the uris of resources (perhaps when the rights are granted via a different deal?). I think (but haven't confirmed) that this already bites us.

This looks great and I look forward to testing it.  Speedup are very welcome.

PR #190, opened 2025-04-06
WIP: fix: return all artists for tracks and albums

sapristi:
Return all artists for tracks and albums. 

Still WIP:
- made the tests pass, although maybe I could improve fixes a bit
- I'm fully sure what's going on, but it seems that on some occasions the artist order is not the same one as returned by tidal


tehkillerbee:
Thanks for your PR. Will continue this work and add it to the next release with some minor changes :+1:

PR #195, opened 2025-06-27
[fix] Don't break tracks import if the track_id is not found

blacklight:
This is rare, but I've had at least a few occurrences of this:

```
ERROR    2025-06-27 16:39:35,962 [235480:MainThread] mopidy.__main__
  Extension tidal failed during setup. This might have left the registry in a bad state.
Traceback (most recent call last):
  File "/usr/lib/python3.13/site-packages/mopidy/__main__.py", line 129, in main
    extension.setup(registry)
    ~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/mopidy_tidal/__init__.py", line 44, in setup
    from .backend import TidalBackend
  File "/usr/lib/python3.13/site-packages/mopidy_tidal/backend.py", line 16, in <module>
    from mopidy_tidal import context, library, playback, playlists
  File "/usr/lib/python3.13/site-packages/mopidy_tidal/library.py", line 466
    track = next(t for t in tracks if t.id == int(track_id))  # pragma: no cover
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
StopIteration
```

It seems to happen when a playlist's contains stale cache entries, and the associated tracks have been removed from Tidal.

### DIFF
--- a/mopidy_tidal/library.py
+++ b/mopidy_tidal/library.py
@@ -462,8 +462,12 @@ class TidalLibraryProvider(backend.LibraryProvider):
         tracks = self._get_album_tracks(session, album_id)
         # If album is unavailable, no tracks will be returned
         if tracks:
-            # We get a spurious coverage error since the next expression should never raise StopIteration
-            track = next(t for t in tracks if t.id == int(track_id))  # pragma: no cover
+            track = next((t for t in tracks if t.id == int(track_id)), None)
+            if not track:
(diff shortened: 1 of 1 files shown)


blacklight:
@tehkillerbee if it's ok for you I'll proceed to merge this one - the use-case is quite straightforward, but I can't write an equally straightforward test for it on top of my mind without mocking the world :)

PR #196, opened 2025-07-05
Support Mopidy v4.0 tracks

blacklight:
The `library.lookup` method should now support both the cases where the tracks returned by Mopidy are Tracks/TlTracks (Mopidy < 4.0) and where they are dict-like representations (Mopidy >= 4.0).

### DIFF
--- a/mopidy_tidal/library.py
+++ b/mopidy_tidal/library.py
@@ -3,7 +3,7 @@ from __future__ import unicode_literals
 import logging
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import suppress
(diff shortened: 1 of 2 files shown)


tehkillerbee:
Thanks for the PR. I did not notice that it was ready for review already, I thought you were still working on it.. :+1:

blacklight:
> Are you planning any other additions? Otherwise, I will prepare for a new release, together with the latest tidalapi.

That'd be great!

I think that these changes are ready - I did some extensive tests in these days with Mopidy 4.0 and nothing broke.

PR #210, opened 2025-12-28
dev: lint with ruff

2e0byo:
This PR sets up to lint with `ruff`.

I'm hestitant to move tooling again (it's the old joke about python), but I've been using `ruff` and `uv` for several years now at work and I really do think they're both a net improvement and here to stay.

This PR just replaces black / isort with ruff, and cleans up all the fluff ruff found (dead code mostly, but also the odd pointless f-string and the like).

If you've not used it, ruff is *very* fast (milliseconds: really), and very easy to install, so I really think we should use it (possibly turning on more lints if needed going forward).

I've added a ci flow to test lints.


2e0byo:
(this was a duplicate)

PR #212, opened 2025-12-28
dev: swap poetry for uv.

2e0byo:
This PR also swaps tooling... this time `poetry` for `uv`. Again, I'm hesitant about this one and think it might merit more discussion, hence keeping it separate (this PR is a strict superset of my ruff PR, so that one can go in separately).

The argument for switching to yet *another* packaging tool is that, again, I've used it for ages professionally and `uv` is quite unlike any other python package manager I've used. It's ridiculously fast: often as much as 100x faster than poetry for resolving deps, and it seems to have taken the python world by storm. Much as I don't really like python these days and would rather write rust, uv *almost* makes me like it again.

I understand if you'd rather stick with poetry, but I really do recommend checking  out this branch and making a few changes (`uv add foo`, `uv lock`, `uv sync`) to see just how much nicer it is.

(This PR *also* sets up the repo to work on nixos, in a purely opt-in way, since that's now my os of choice. Getting that to work with gstreamer for local dev is my *actual* motivation for uv here, and the intended work is coming in the next PR.)

### DIFF
--- a/mopidy_tidal/playback.py
+++ b/mopidy_tidal/playback.py
@@ -38,19 +38,18 @@ class TidalPlaybackProvider(backend.PlaybackProvider):
                 )
 
         stream = session.track(track_id).get_stream()
-        manifest = stream.get_stream_manifest()
         logger.info("MimeType:{}".format(stream.manifest_mime_type))
-        logger.info(
-            "Starting playback of track:{}, (quality:{}, codec:{}, {}bit/{}Hz)".format(
-                track_id,
-                stream.audio_quality,
-                manifest.get_codecs(),
-                stream.bit_depth,
-                stream.sample_rate,
-            )
-        )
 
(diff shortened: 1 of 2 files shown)


2e0byo:
(test failure again is just codecov rate limiting. I think you mange the account @tehkillerbee : if you get a moment would you be able to copy a token and put it in a repo / org secret? then we could pull it in and avoid the rate limit. But that's def low prio.)

tehkillerbee:
I did consider using uv as well for a project related to work. However, I have not yet looked into it.

I have no issue switching to uv, although I prefer if we then use the same package management tool for python-tidal. Is this difficult to implement for python-tidal as well?

tehkillerbee:
One thing I am not quite sure about is that we still have a bit of a mix of uv and poetry in the repository now. Is this intentional?

PR #213, opened 2025-12-28
feat: playback proxy

2e0byo:
This PR adds an http proxy for playback suitable for use over slow / unreliable connections. Unlike my previous proxy from years ago, this is a fully-featured (if minimal) http proxy sitting between gstreamer and tidal.

It works as follows:

- The proxy runs in a thread, running a python asyncio loop (for easier concurrency)
- gstreamer is given a local (proxy) url
- if the proxy has data in its cache it serves the cached data. Otherwise it streams data from the remote into the local cache and out to gstreamer.
  - if the stream finishes successfully, the cached data is finalised and will be used next time. If the stream is dropped (e.g. by the user skipping tracks before buffering is finished), the cached data is dropped.
  
Note that caching is only applied when mpeg-DASH is not used, since caching an adaptive stream makes no sense (every chunk in mpeg-DASH is like a little file with one of multiple encodings adapting to network conditions. We *could* make this work by just retrying until we get the high-quality section, but that really defeats mpeg-DASH. Naturally, this means the cache won't work with formats requiring mpeg-DASH. I don't think that's a problem, as you can't listen to ultrasonic encoded audio over 4G anyway ;).

The work is:

- [X] implement http cache
- [x] support https if required
- [x] plug into mopidy-tidal's url transformer
- [x] add cache eviction
- [ ] <strike> add an offline download api (this will never be supported in mopidy as AKAIK it has no such context, but we can ship a tiny cli so you can pull your playlists before going on a trip).</strike> I have a POC download command, but it's sufficiently complicated it'll have to land in another pr
- [ ] consider getting this to work with pkce. This would require some way to get a real download url from tidalapi, rather than a stream manifest. @tehkillerbee do you  remember if this is possible?

The proxy works  fully offline if you have played a track recently, since:

- the uri -> remote url conversion is cached in the database
- the remote url is then cached in the database
- we already cache all the other tidal resources

If you use this with lazy login, you can take your laptop off the wifi connection and it will have a fair chance of continuing to play --- if your listening habits are like mine :D

Note that using a DB for the cache made a true LRU cache very easy to implement. I wonder if we might consider swapping it in for the other caches? This would have the advantage of putting eviction in one place, i.e. if you evict the track metadata row you also evict the track, etc. Plus it wouldn't be FIFO ;)

I'm in no rush to get this in though obviously I hope it's something we could use more widely. Right now I'm using this branch on all my systems and testing it out.

The history is a bit of a mess: I've squashed a fair bit of it today but it's still a record of me getting things wrong  rather than neatly added features. If that's a problem I can do a better squash, or we can squash the whole thing. 

For the fun of it this is a zero-dep pure python impl.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,59 @@ sudo pip3 install .
 sudo systemctl restart mopidy
 ```
 
+### Install / develop with nix
+
+If you are running nix, you can drop into a shell with everything ready to go
+using
+
+```shell
+nix develop
+```
+
+or, with `direnv` installed,
+
+```shell
+direnv allow
+```
+
+You should *not* attempt to install the `complete` group: instead nix is used to
+provide the needed non-python libraries (and a few extensions for manual and
+automated testing).  If you are not running any other mopidy on your system, the
+integration tests should pass.
+
+Mopidy-Tidal is in nixpkgs, but to use the latest git or a local checkout you
+can override to build directly from this flake.  Assuming you use flakes, the
+easiest approach is to add a new entry to `inputs`:
+
+```nix
+inputs.mopidy-tidal = {
+  url = "git+file:///home/USER/code/mopidy-tidal"; # or github:EbbLabs/mopidy-tidal
+  inputs.nixpkgs.follows = "nixpkgs";
+};
+```
+
+Then (making sure to pass `inputs` through) you can do something like:
+
+```nix
+  services.mopidy = let
+    system = pkgs.stdenv.hostPlatform.system;
+    mopidy-tidal = inputs.mopidy-tidal.outputs.packages.${system}.default;
+  in {
+    enable = true;
+    extensionPackages = with pkgs;
+      [mopidy-local mopidy-iris mopidy-mpd mopidy-mpris]
+      ++ [mopidy-tidal];
+    settings = {
+      local.media_dir = "~/Music/Library";
+      local.scan_follow_symlinks = true;
+      tidal.quality = "LOSSLESS";
+      tidal.playback_cache = true;
+    };
+  };
+```
+
+substituting for your own local configuration as appropriate.
+
 ## Dependencies
 
 ### Python


2e0byo:
The remaining issue with this approach is seeking, which it was designed to solve. Seeking works fine if the track is buffered into the cache, but if not gstreamer opens a new range request (which is perfectly correct), but unfortunately the sql cache isn't concurrent safe. My fault for forgetting that sqlite isolation is per-connection.

Once that works I want to test seeking fairly thoroughly, since that's really what makes or breaks a "transparent" cache. Then the cache should be re-worked slightly to have an optional cache key which isn't the url, and we can use that to avoid needing to hit the network at all until we get a cache miss (I went with the url's path as the key for now just to get things up and running).

Sticking this all here as I'm back to work tomorrow and don't know how much further I'll get, but I'm actively using this for playback atm.

2e0byo:
fixed partial insertion.

2e0byo:
With this push everything is implemented except the pkce + offline download. I have a POC of the offline download and it's going to be sufficiently complicated I'll land it in a separate PR, either into this branch or upstream. For PKCE I'll wait for your input, but right now we could just document this as only working with oauth and look into it later.

Changes:

- it seems we *can* use a thread, not a process, to run the cache. An earlier lockup which I was convinced was the GIL appears to be something inside gstreamer itself. On nixos with this flake I can't play urls until I've first played a local file (? but my problem, not yours).
- the cache DB now caches uri -> path conversion. This allows the proxy to continue caching at the HTTP level, whilst removing the need for the network when looking up a track we have cached. This in turn permits fully offline music playing for cached music. I'd like to make mopidy-tidal more offline aware full stop (currently it's a bit of a lottery if you'll hit a path requiring login, although most paths will just error and skip).

Naturally you might ask if the cache shouldn't just work with uris all along. I don't think that's a bad way to build it, but I've gone for the other way now :D. I did consider appending a custom http query param like `_mopidy_tidal_uri=tidal:track:0:0:0` to the generated uri and having the cache strip it before hitting upstream, but decided this is cleaner.

2e0byo:
Drat we do need to add a a new config option for the cache max size and pass it through.

DEVELOPMENT.md
tehkillerbee: This file still has several references to poetry etc. Shouldn't that be updated too?

2e0byo: oops, ta. Will clean up.

tehkillerbee:
> do you remember if this is possible?

If you mean downloading and assembling into an actual audio data, this is possible but requires some extra work. IIRC, this is what the tidal-dl does.

integration_tests/test_login_hack.py
```diff
@@ -17,20 +21,25 @@
...
-def test_user_warned_if_lazy_set_implicitly(spawn, config_dir):
-    config = config_dir / "hack.conf"
-    with spawn(f"mopidy --config {config.resolve()}") as child:
-        child.expect("Connecting to TIDAL... Quality = LOSSLESS")
-        child.expect("HACK login implies lazy connection")
+def test_user_warned_if_lazy_set_implicitly(spawn, config_dir: Path, tmp_path: Path):
+    with spawn(mopidy(config_dir, "hack.conf", tmp_path)) as child:
+        child.expect("Quality: LOSSLESS")
+        child.expect("Authentication: OAuth")
+        child.expect("implies lazy connection")
```
tehkillerbee: Should this line be capitalized? Or is there a missing word?

mopidy_tidal/backend.py
```diff
@@ -54,6 +55,22 @@ def __init__(self, config, audio):
...
         self.login_server_port: int = 8989
 
+    @property
+    def track_cache(self) -> ThreadedProxy | None:
+        path = (
+            Path(
+                self._tidal_config["track_cache"]
+                and Extension.get_cache_dir(self._config)
+            )
+            / "track.db"
```
tehkillerbee: Can the track db grow very large over time? Or do we limit it to a number of entries, as the config param hints?

mopidy_tidal/playback.py
```diff
@@ -1,8 +1,13 @@
 from __future__ import unicode_literals
 
 from typing import TYPE_CHECKING
+from urllib.parse import urlparse
 
+import tidalapi
 from login_hack import speak_login_hack
```
tehkillerbee: We should probably rename the login_hack at some point :)

(Not related to this MR of course...)

mopidy_tidal/playback.py
```diff
@@ -19,61 +24,86 @@
...
+    if stream.manifest_mime_type == ManifestMimeType.MPD:
+        logger.info(
+            "Starting playback of track:{}, (quality:{}, {}bit/{}Hz)".format(
+                track.id,
+                stream.audio_quality,
+                stream.bit_depth,
+                stream.sample_rate,
+            )
+        )
+        data = stream.get_manifest_data()
```
tehkillerbee: The manifest data themselves (i.e. individual stream URLs) can probably be cached directly, using the same approach as used here) for instance using StreamManifest => get_urls()

tehkillerbee (approved):
LGTM! If possible, please add some more docs/comments to the new classes added.

mopidy_tidal/backend.py
2e0byo: it's limited if the config is set, unlimited by default. We could pick a sensible default e.g. 1k rows?

integration_tests/test_login_hack.py
2e0byo: it's actually a partial regex match equivalent to ".*implies lazy connection.*".

2e0byo:
Still needs some user docs and should be rebased against main when the other nix pr is in.

@tehkillerbee is there anything else you'd like documented / could be clearer? it's never easy documenting your own code...

tehkillerbee:
> Added docs. I'd rather the nix PR went in first if that's alright and I'll rebase this one on top? (I have a branch local ready to go). That is if you're happy with the latest changes.

Perfect, I have merged that one now :+1:

2e0byo:
Ok, this is ready to go; I added some docs on nix to help with things.

I'm pretty sure there are still improvements to be made on the caching performance esp for small systems, but those can come later.

2e0byo:
Just rebased against main.

tehkillerbee:
Merging... Thanks for your contributions :1st_place_medal:

PR #238, opened 2026-04-28
fix/proxy remote url can change

2e0byo:
When I wrote the proxy, I assumed tidal's upstream URL would stay fixed.  Here on holiday in Milan, it's changed.  I don't know if that's regional or not, but it makes sense to look it up.

I've assumed that all tracks will have the same proxy url within a session.  This could still fail if:

- you take a long running mopidy session over a national border (if it's regional)
- different tracks have different remote proxy urls

In which case let's open an issue and fix it when it comes in.

The `Lazy[T]` proxy class is my current favourite way to get a typesafe `| None`.

### DIFF
--- a/mopidy_tidal/backend.py
+++ b/mopidy_tidal/backend.py
@@ -5,6 +5,7 @@ import time
 from concurrent.futures import Future
 from pathlib import Path
 from typing import Optional, Union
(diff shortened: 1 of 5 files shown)


tehkillerbee:
LGTM!

PR #239, opened 2026-04-28
refactor/type safe session

2e0byo:
This PR makes the `._active_session` attr type safe, i.e. gets rid of all the red underlines from your IDE telling you it might be None.

This isn't IMHO the best way to handle this kind of state, but it's quick to do and should make the dev experience nicer.  I could easily have done the others, I just wanted to show what it would look like.  Open to thoughts.

[This PR is a superset of #238]


tehkillerbee:
LGTM, I think this is an improvement.

PR #240, opened 2026-04-30
perf/blob io

2e0byo:
This PR is on top of #239, though if that's rejected it will rebase on top of #238.

It replaces multiple chunk-sized records with a single record for the body, allocated up front with the correct size, and then buffered with sqlite's blobopen function (which returns a file pointer to a blob record).  I've kept it compatible with the old format and it continues to play cached data in that format.

This should lead to improved DB performance, as previously the DB fragmented very quickly, and hopefully will lead to reduced RAM usage.

I've also taken the opportunity to clean up dead / redundant code from the development process of the proxy.

From using the cache for several months I've noticed:

- RAM usage sometimes kills mopidy on my raspberry pi (the ram headroom is marginal anyway and I'm running a pretty hevayweight NixOS I should tune, but it's definitely cache related)
- if gstreamer crashes, pykka can re-spawn a thread with a now stale DB connection

Gstreamer crashes (based on seeking around aggressively) seem to go away when I used the default HTTP sink and not curl.

### DIFF
--- a/mopidy_tidal/backend.py
+++ b/mopidy_tidal/backend.py
@@ -15,6 +15,7 @@ from tidalapi import __version__ as tidalapi_ver
 from mopidy_tidal import Extension, context, library, playback, playlists
 from mopidy_tidal import __version__ as mopidy_tidal_ver
 from mopidy_tidal.gstreamer_proxy import ThreadedProxy, mopidy_playback_cache
+from mopidy_tidal.types import Lazy
 from mopidy_tidal.web_auth_server import WebAuthServer
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


tehkillerbee (approved):
LGTM! 

By the way, does it make sense to increase the default cache size to a size larger than 16MB?

2e0byo:
re cache (buffer) size... I run on a rPi 0 hence being very conservative (playing a large enough album and seeking aggressively is still enough to kill it, though the base OS is responsible for about half the memory usage and I need to tune it).

Larger sizes are basically free on desktop, but the "proxy" is serving over localhost anyway and just needs to be faster than playback.  I've not profiled though and python's memory management is *horrible*.

Re. the last point in the first message (unaddressed here): under very high contention there's still some behaviour I don't fully understand  in pykka where it will spawn a new worker thread and they'll then conflict on the DB (I'm guessing it has some kind of unresponsive timeout for deciding a worker is dead, since I'm pretty sure nothing actually died).  At some point I'll give in and just catch the error and re-open the DB, but again I can only trigger it on a rpi which has basically run out of ram by that point.

PR #243, opened 2026-05-11
Add HTTP API for managing TIDAL favorites

### LINKED ISSUE #241: Add Tidal favorites support (add/remove/list albums, tracks, artists)
---
  Title: Add Tidal favorites support (add/remove/list albums, tracks, artists)

  Body:

  ## Summary

  `tidalapi` already exposes the favorites API (`session.user.favorites.add_album()`,
  `remove_album()`, `albums()`, plus the equivalents for tracks, artists, playlists),
  but `mopidy-tidal` doesn't surface it. This means clients can only browse
  favorites that were marked elsewhere (web app, official client) — they can't
  favorite/unfavorite from a Mopidy client.

  ## Use case

  Native Mopidy clients (Iris, custom UIs, mopyrust, etc.) want to offer a
  "favorite this album" button that adds the album to the user's Tidal library so
  it shows up under `tidal:my_albums` on the next browse — without each client
  re-implementing OAuth and reusing the session that `mopidy-tidal` has already
  authenticated.

  ## Proposed API

  I see two reasonable shapes — happy to take direction:

  **Option A — extend the existing HTTP module** (preferred; mirrors how the
  OAuth/image-proxy endpoints already work):

  POST   /tidal/favorites/albums   { "id": "12345" }
  DELETE /tidal/favorites/albums/12345
  GET    /tidal/favorites/albums

  …with the same shape for `/tracks`, `/artists`, `/playlists`. Authentication is
  implicit (uses the session already loaded by the backend).

  **Option B — abuse Mopidy's playlists API** by exposing favorites as a synthetic
  playlist (`tidal:favorites:albums`). Probably worse because favorites aren't
  ordered playlists and Mopidy's playlist model doesn't fit cleanly.

  ## Implementation sketch

  Roughly:

  ```python
  # mopidy_tidal/web.py (existing module)
  class FavoritesHandler(RequestHandler):
      def initialize(self, session):
          self._session = session

      def post(self, kind):
          body = json.loads(self.request.body)
          getattr(self._session.user.favorites, f"add_{kind}")(body["id"])
          self.set_status(204)

      def delete(self, kind, id):
          getattr(self._session.user.favorites, f"remove_{kind}")(id)
          self.set_status(204)

  …wired into the existing app factory next to the OAuth/image routes.

  Happy to send a PR if you're open to it — wanted to check direction first
  since the API shape (HTTP vs. extension hook vs. something else) is your call.

  ---

prietus:
## Summary

Implements the favorites HTTP API discussed in #241 (option A — endpoints in
the existing HTTP frontend).

- Registers a Mopidy `http:app` so the routes mount at `/tidal/` alongside
  the rest of the Mopidy HTTP frontend.
- `POST /tidal/favorites/<kind>s` with body `{ "id": "<id>" }` adds a
  favorite. Returns 204.
- `DELETE /tidal/favorites/<kind>s/<id>` removes a favorite. Returns 204.
- `GET /tidal/favorites/<kind>s` lists current favorites as
  `[{ id, name, artist? }, ...]`.
- `<kind>` is one of `album`, `track`, `artist`, `playlist`.
- Authentication is implicit: the handlers reach the `TidalBackend` actor
  via `pykka.ActorRegistry` and reuse the `tidalapi.Session` it already
  holds, so clients don't re-implement OAuth.
- Returns 503 if the backend isn't logged in yet, 400 for invalid/empty body
  on POST.

Closes #241.

## Test plan

- [x] `uv run pytest tests/test_web.py tests/test_extension.py` — 31 passing
- [x] `uv run ruff format --check` — clean
- [x] `uv run ruff check` — clean
- [ ] Smoke-tested against a live Mopidy + mopidy-tidal: POST adds the album
      to the TIDAL library, GET lists it back, DELETE removes it, and the
      changes show up on the official client / web app on next refresh.

### DIFF
--- a/mopidy_tidal/web.py
+++ b/mopidy_tidal/web.py
@@ -1,0 +1,120 @@
+"""HTTP endpoints for managing the user's TIDAL favorites.
+
+Mounted under ``/tidal/`` by Mopidy's HTTP frontend. Routes:
+
+    POST   /tidal/favorites/<kind>s          { "id": "12345" }
+    DELETE /tidal/favorites/<kind>s/<id>
+    GET    /tidal/favorites/<kind>s
+
+where ``<kind>`` is one of ``album``, ``track``, ``artist``, ``playlist``.
+
+Authentication is implicit: the handlers reuse the ``tidalapi.Session`` already
+held by the ``TidalBackend`` actor, so clients don't re-implement OAuth.
+"""
+
+from __future__ import unicode_literals
+
+import json
+import logging
+
+import pykka
+from tornado.web import HTTPError, RequestHandler
+
+logger = logging.getLogger(__name__)
+
+# Tidal entity kinds we expose. Each must map to ``tidalapi.Favorites`` having
+# methods ``add_<kind>``, ``remove_<kind>`` and an attribute ``<kind>s``.
+KINDS = ("album", "track", "artist", "playlist")
+
+
+def make_app(config, core):
+    """Mopidy ``http:app`` factory. Looks up the running TidalBackend actor
+    and binds it into every handler so requests can reach the session."""
+
+    # Late import: backend.py imports from this package at module load and
+    # would cause a circular import otherwise.
+    from mopidy_tidal.backend import TidalBackend
+
+    refs = pykka.ActorRegistry.get_by_class(TidalBackend)
+    if not refs:
+        # Should not happen: this factory is only invoked when the extension
+        # is enabled, which also instantiates the backend.
+        raise RuntimeError(
+            "TidalBackend actor not running; cannot register HTTP routes"
+        )
+    backend = refs[0].proxy()
+
+    common = {"backend": backend}
+    return [
+        (
+            r"/favorites/(album|track|artist|playlist)s",
+            FavoritesCollectionHandler,
+            common,
+        ),
+        (
+            r"/favorites/(album|track|artist|playlist)s/([^/]+)",
+            FavoritesItemHandler,
+            common,
+        ),
+    ]
+
+
+class _Base(RequestHandler):
+    def initialize(self, backend):
+        self.backend = backend
+
+    def write_error(self, status_code, **kwargs):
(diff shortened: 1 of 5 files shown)


mopidy_tidal/web.py
```diff
@@ -0,0 +1,120 @@
...
+    from mopidy_tidal.backend import TidalBackend
+
+    refs = pykka.ActorRegistry.get_by_class(TidalBackend)
+    if not refs:
+        # Should not happen: this factory is only invoked when the extension
+        # is enabled, which also instantiates the backend.
+        raise RuntimeError(
+            "TidalBackend actor not running; cannot register HTTP routes"
+        )
+    backend = refs[0].proxy()
```
2e0byo, line 45: what is this proxy object? a pykka proxy, right?

2e0byo:
(I can't approve the workflow.)  From a quick glance the code looks good.

I'd question the choice of 503, although it is defensible.

> The 503 (Service Unavailable) status code indicates that the server is currently unable to handle the request due to a temporary overload or scheduled maintenance, which will likely be alleviated after some delay. [RFC9110]

Neither of these is true here: it's just that mopidy-tidal is using lazy login and we need you to login. I think you should definitely make sure the error text says something like "please ensure that you can play music from tidal using mopidy-tidal and then retry", and consider raising 403 instead, although the re-auth flow is a bit weird ("go and play a track").

Thanks for the code, this is definitely something it'd be good to have.

prietus:
Thanks for the review! Based on the feedback (and given the
  workflow-approval issue) I'm going to keep this living in a companion package (github.com/prietus/mopidy-tidal-goodies)
  rather than push for upstream inclusion — that lets me iterate at my own cadence and keeps mopidy-tidal's scope tighter.

  Applied the 503→403 distinction you suggested there (with an actionable error message pointing at the lazy-login flow).
  Closing this and deleting the branch.

  Thanks again for the time.

